PR #8, opened 2019-03-13
REG-82 fix overflowing bar charts and style errors

emro:
Previously when you would select an annotation term to filter results, the bars on the other annotation terms would overflow.
This resulted from two facets being combined ("assay_term_name" and "annotation_type"). Now, the facets are not combined, they are just designed to look like they are combined, and this resolves the issue.
Other updates on this branch are to fix linting errors.

### DIFF
--- a/src/encoded/static/components/region_search.js
+++ b/src/encoded/static/components/region_search.js
@@ -7,7 +7,7 @@ import { Panel, PanelBody } from '../libs/bootstrap/panel';
 import { FacetList, FilterList } from './search';
 import { SortTablePanel, SortTable } from './sorttable';
 
-const DataTypeStrings = [
+const dataTypeStrings = [
     {
         type: 'dbSNP IDs',
         explanation: 'Enter dbSNP ID(s) (example) or upload a list of dbSNP IDs to identify DNA features and regulatory elements that contain the coordinate of the SNP(s).',
@@ -22,7 +22,7 @@ const DataTypeStrings = [
     },
 ];
 
-const ExampleEntries = [
+const exampleEntries = [
     {
         label: 'multiple dbSNPs',
         input: 'rs3768324\nrs75982468\nrs10905307\nrs10823321\nrs7745856',
@@ -46,9 +46,9 @@ class DataType extends React.Component {
     }
 
     toggleOpenState() {
-        this.setState((state) => {
-            return { open: !state.open };
-        });
+        this.setState((state) => ({
+            open: !state.open
+        }));
     }
 
     handleInfo() {
@@ -152,10 +152,10 @@ class AdvSearch extends React.Component {
 
                                 <div className="example-inputs">
                                     Click for example entry:
-                                    {ExampleEntries.map((entry, entryIdx) =>
+                                    {exampleEntries.map((entry, entryIdx) =>
                                         <span key={entry.label}>
                                             <ExampleEntry label={entry.label} input={entry.input} handleExample={this.handleExample} />
-                                            { entryIdx !== (ExampleEntries.length - 1) ?
+                                            { entryIdx !== (exampleEntries.length - 1) ?
                                                 'or'
                                             :
                                             null}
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/search.js
```diff
@@ -1050,10 +1041,9 @@ export class FacetList extends React.Component {
         // are the negation facet terms that need to get merged into the regular facets that their
         // non-negated versions inhabit.
         const negationFilters = filters.filter(filter => filter.field.charAt(filter.field.length - 1) === '!');
-        
-        console.log("building facets");
-        
-        console.log(width);
+
+        // If there are no assay terms, the annotation type facet will need the "Method" label
+        const assayTerms = facets.filter(facet => facet.field === 'assay_term_name').some(facet => facet.terms.some(f => f.doc_count > 0));
```
forresttanaka: Could this use `.find()` instead of `.filter()`? Or can there be more than one assay_term_name facet?

If you can use `.find()`, you’ll probably need to break this into two or three lines in case `.find()` doesn’t find anything, but that might be good because I’m having a tough time wrapping my head around this line. I _think_ that way you would only need one `.some()`.

src/encoded/static/components/search.js
```diff
@@ -974,39 +977,27 @@ TextFilter.propTypes = {
...
-                    facetsModified[0].terms.push(f);
-                }
-            });
-        
-            facetsModified[0].title = 'Method';
-            facetsModified[1].title = 'Biosample';
+
+        const facetsModified = facets;
+        if (modifyFacetsFlag) {
+            const methodIndex = facetsModified.map(facet => facet.title).indexOf('Assay');
```
forresttanaka: Could change this from .map().indexOf() to [.findIndex()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) so it can be done in one step, and also .findIndex() stops after it finds a match where .map() goes through the whole array no matter what.

src/encoded/static/components/search.js
```diff
@@ -1050,10 +1041,12 @@ export class FacetList extends React.Component {
...
-        
-        console.log("building facets");
-        
-        console.log(width);
+
+        // If there are no assay terms, the annotation type facet will need the "Method" label
+        // First we will check to see if there is an "assay_term_name" facet
+        const assayFacet = facets.find(facet => facet.field === 'assay_term_name');
+        // Then we will check to make sure that there is at least one "assay_term_name" term with "doc_count" greater than 0
+        const assayTerms = assayFacet ? assayFacet.terms.some(f => f.doc_count > 0) : false;
```
forresttanaka: Not that it makes a huge/any difference in performance, but you can shorten this to:

```
const assayTerms = assayFacet && assayFacet.terms.some(f => f.doc_count > 0);
```

If assayFacet is undefined, with this method you won’t get “false” but you will get falsy.

src/encoded/static/components/region_search.js
```diff
@@ -83,53 +85,43 @@ class AdvSearch extends React.Component {
...
-                                </textarea>
+                                <textarea className="multiple-entry-input" id="multiple-entry-input" placeholder="Enter search parameters here." onChange={this.handleChange} name="regions" value={this.state.searchInput} />
 
-                                <p className="example-inputs" onClick={this.handleExamples}>Click for example entry: <span className="example-input" id="example-snps">multiple dbSNPs</span> or <span className="example-input" id="example-coordinates">coordinates ranges</span></p>
+                                <p className="example-inputs">
+                                    Click for example entry: <span className="example-input" id="example-snps" onClick={e => this.handleExamples(e)} onKeyDown={this.handleExamples} role="button" tabIndex={0}>multiple dbSNPs</span> or <span className="example-input" id="example-coordinates" onClick={e => this.handleExamples(e)} onKeyDown={this.handleExamples} role="button" tabIndex={0}>coordinates ranges</span>
```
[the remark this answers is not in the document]
emro: I tried to write this so that each example was a component, and it works, except it also submits the input by triggering handleOnFocus, which is confusing and not what we want. Can you take a look at that?

src/encoded/static/components/region_search.js
```diff
@@ -1,66 +1,102 @@
...
+import { Panel, PanelBody } from '../libs/bootstrap/panel';
+import { FacetList, FilterList } from './search';
 import { SortTablePanel, SortTable } from './sorttable';
 
-
-const regulomeGenomes = [
-    { value: 'GRCh37', display: 'hg19' },
-    { value: 'GRCh38', display: 'GRCh38' },
+const DataTypeStrings = [
+    {
```
forresttanaka: You need to change `DataTypeStrings` to camel case, along with `ExampleEntries`. Didn’t know until now AirBnB has a rule on that (which ESLint didn’t pick up): https://github.com/airbnb/javascript#naming--camelCase

PR #13, opened 2019-04-09
REG-92 fix clickable FAQ for Safari and Firefox

emro:
I was detecting which question a user had clicked on using a method only available in Chrome. This solution should be compatible with all browsers!

### DIFF
--- a/src/encoded/static/components/blocks/richtext.js
+++ b/src/encoded/static/components/blocks/richtext.js
@@ -67,7 +67,6 @@ export default class RichTextBlockView extends React.Component {
                     // Polyfill for 'closest' for IE
                     const elementPrototype = window.Element.prototype;
                     if (!elementPrototype.closest) {
-                        console.log('using the polyfill');
                         elementPrototype.closest = function(s) {
                             let el = this;
                             do {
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/blocks/richtext.js
```diff
@@ -58,32 +57,46 @@ export default class RichTextBlockView extends React.Component {
...
-                        target = e.path[i];
-                        break;
+        if (this.state.value.body.includes('faq')) {
+            const faqContainer = document.getElementsByClassName('faq')[0];
+            if (faqContainer) {
+                // One event listener on the 'faq' container handles all click events
+                faqContainer.addEventListener('click', (e) => {
+                    // Polyfill for 'closest' for IE
+                    const elementPrototype = window.Element.prototype;
+                    if (!elementPrototype.closest) {
```
forresttanaka: This test for `closest` struck me as a little bit odd because this test would happen every time you click. But I think it’s OK since the performance penalty would be probably unmeasurable. But I think it’s more expected to test and assign the polyfill in componentDidMount().

emro: Yes, that would be better. I was actually sort of wondering if it should be in objectutils.js or app.js. If we wanted to use `closest` anywhere else it wouldn't make sense to keep it here.

forresttanaka: Maybe even browserfeat.js? That only tests for features now, but it could add features.

PR #14, opened 2019-04-15
REG-94 IE bug fixes

emro:
Fixes two bugs on IE:
(1) /regulome-help/ is now a proper link, not a link in a button
(2) Bar chart annotations are now right-aligned on IE


src/encoded/static/scss/encoded/modules/_regulome_search.scss
```diff
@@ -531,6 +535,7 @@ table {
 
 .facet-term__text {
     max-width: 42%;
+    flex: 1;
 }
```
emro: This is the line that fixed the alignment.

PR #15, opened 2019-04-16
REG-90 Encode-related updates

emro:
Encode has some updates which we need to sort out in Regulome. We need to:
- Remove cart feature
- Add typeahead facets
- Sort numerical facets by value not by frequency
- Change University of Michigan logo to be blue
- Avoid console errors for special characters in typeahead search bars


hitz:
@forresttanaka you can skip everything but the JS changes to review this.

emro:
I considered taking out the cart more thoroughly but thought there was a very small chance we could want to bring it back and didn't think it would hurt performance to keep some of the scripts.

PR #18, opened 2019-04-30
REG-102 remove cart test

emro:
We removed the cart from Regulome so we do not want the cart tests.


yunhailuo:
Ready once tests pass.

emro:
Do I need to build a demo for this one, since it just changed the tests, which we ran?

yunhailuo:
> Do I need to build a demo for this one, since it just changed the tests, which we ran?

No need. Good to go.

PR #19, opened 2019-05-01
REG-98 update regulome facets

yunhailuo:
Genome assembly has to in facet (for elastic aggregation) at the backend because it's used later for making visualization links. We can change that behavior later when we got other visualization tools.

### DIFF
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -325,7 +325,7 @@ class RegulomeSearch extends React.Component {
         const notification = context.notification;
         const searchBase = url.parse(this.context.location_href).search || '';
         const filters = context.filters;
-        const facets = context.facets;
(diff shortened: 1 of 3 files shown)


yunhailuo:
Rebased to dev after #20 by force-push.

src/encoded/static/components/regulome_search.js
```diff
@@ -325,7 +325,7 @@ class RegulomeSearch extends React.Component {
         const notification = context.notification;
         const searchBase = url.parse(this.context.location_href).search || '';
         const filters = context.filters;
-        const facets = context.facets;
+        const facets = context.facets.filter(facet => facet.title !== "Genome assembly");;
```
emro: Only need one semicolon

yunhailuo: Thanks. Really stupid. Fixed. Sorry.

hitz:
This is going to get replaced by new UI.

PR #21, opened 2019-05-02
REG-71 add PWM logos

emro:
*This is a new PR because the last one had several issues, but it is largely the same.*

Adding motifs for PWMs.
Eventually we will want some back end changes so that we don't display duplicate logos, but this is a good start.
We are linking the d3-sequence-logo library from GitHub but we have a local version which was useful for faster development.

### DIFF
--- a/src/encoded/static/components/motifs.js
+++ b/src/encoded/static/components/motifs.js
@@ -41,104 +41,119 @@ export function getMotifData(pwmLink, fetch) {
 }
 
 // Display information on page as JSON formatted data
-export class Motifs extends React.Component {
+export class MotifElement extends React.Component {
     constructor() {
         super();
 
-        this.drawMotifs = this.drawMotifs.bind(this);
+        // this.drawMotifs = this.drawMotifs.bind(this);
         this.addMotifElement = this.addMotifElement.bind(this);
     }
 
     componentDidMount() {
         require.ensure(['d3', 'd3-sequence-logo'], (require) => {
-            if (this.chartdisplay) {
-                this.d3 = require('d3');
-                this.sequenceLogos = require('d3-sequence-logo'); // logos This is for local development when changes are needed to d3-sequence-logo.
-                const targetElement = this.chartdisplay;
-                this.drawMotifs(targetElement);
-            }
+            this.d3 = require('d3');
+            this.sequenceLogos = require('d3-sequence-logo'); // logos This is for local development when changes are needed to d3-sequence-logo.
+
+            const targetElement = this.chartdisplay;
+            const entryPoint = this.sequenceLogos.entryPoint;
+
+            getMotifData(this.pwmLink, this.context.fetch).then((response) => {
+                this.addMotifElement(targetElement, response, entryPoint);
+            });
         });
     }
 
     // Redraw charts when window changes
     componentDidUpdate() {
-        if (this.chartdisplay) {
-            this.chartdisplay.innerHTML = '';
-            const targetElement = this.chartdisplay;
-            this.drawMotifs(targetElement);
-        }
+        this.d3 = require('d3');
+        this.sequenceLogos = require('d3-sequence-logo'); // logos This is for local development when changes are needed to d3-sequence-logo.
+
+        const targetElement = this.chartdisplay;
+        const entryPoint = this.sequenceLogos.entryPoint;
+
+        // Need to remove current logo or another logo will be appended to original
+        this.chartdisplay.innerHTML = '';
+
+        getMotifData(this.pwmLink, this.context.fetch).then((response) => {
+            this.addMotifElement(targetElement, response, entryPoint);
+        });
     }
 
-    addMotifElement(targetElement, response, d, entryPoint) {
+    addMotifElement(targetElement, response, entryPoint) {
         // Convert PWM text data to object
         const PWM = convertTextToObj(response);
-        // Append information about the PWM including Biosample, Organ, and Targets
-        targetElement.insertAdjacentHTML('afterbegin', `<div class="motif-element" id="motif${d.accession}"></div>`);
-        const targetList = d.targets.map(t => t.label);
-        const targetListLabel = targetList.length > 1 ? 'Targets' : 'Target';
-        const htmlStr = `
-            <div class='motif-description'>
-                <p><a href=${d['@id']}>${d.accession}</a></p>
-                ${d.biosample_term_name ? `<p>Biosample: ${d.biosample_term_name}</p>` : ''}
-                ${d.organ_slims ? `<p>Organ: ${d.organ_slims.join(', ')}</p>` : ''}
-                ${targetList ? `<p>${targetListLabel}: ${targetList.join(', ')}</p>` : ''}
-            </div>`;
-        targetElement.insertAdjacentHTML('afterbegin', htmlStr);
         // Generate the logo from the PWM object
-        entryPoint(`#motif${d.accession}`, PWM, this.d3);
-    }
-
-    drawMotifs(targetElement) {
-        const results = this.props.context['@graph'];
-        const urlBase = this.props.urlBase;
-
-        // Keep track of whether or not a PWM document has been identified
-        let emptyFlag = 1;
-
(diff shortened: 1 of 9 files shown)


src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,144 @@
...
+        results.forEach((d) => {
+            if (d.documents[0]) {
+                emptyFlag = 0;
+                if (d.documents[0].document_type === 'position weight matrix') {
+                    pwmLink = urlBase + d.documents[0]['@id'] + d.documents[0].attachment.href;
+
+                    getMotifData(pwmLink, this.context.fetch, entryPoint, d3).then((response) => {
+                        targetElement.insertAdjacentHTML('afterbegin', `<div class="element" id="element${d.accession}"></div>`);
+                        const targetElement2 = document.getElementById(`element${d.accession}`);
+                        this.addMotifElement(targetElement2, response, d, entryPoint, d3);
```
forresttanaka: addMotifElement doesn’t have a last d3 parameter — it uses this.d3.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,144 @@
...
+    // Redraw charts when window changes
+    componentDidUpdate() {
+        if (this.chartdisplay) {
+            this.chartdisplay.innerHTML = '';
+            const targetElement = this.chartdisplay;
+            this.drawMotifs(targetElement);
+        }
+    }
+
+    addMotifElement(targetElement, response, d, entryPoint) {
```
forresttanaka: Without spending the time that you have on this, I could be completely off, but I feel like a lot of direct DOM manipulation could be avoided in this method by using a state set by getMotifData() to render these divs in `render()`, and some of these divs use props, so they wouldn’t even need to use state. The only exception would be entryPoint which couldn’t be called until after mount, but you can get to the divs that it renders into through refs, or through getElementById in the rare case that refs can't handle this.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,144 @@
...
+                        targetElement.insertAdjacentHTML('afterbegin', `<div class="element" id="element${d.accession}"></div>`);
+                        const targetElement2 = document.getElementById(`element${d.accession}`);
+                        this.addMotifElement(targetElement2, response, d, entryPoint, d3);
+                    });
+                }
+            }
+        });
+
+        // If none of the elements of @graph have a PWM document, display notification
+        if (emptyFlag) {
```
forresttanaka: `emptyFlag` is determined purely by props, so this could be done in `render()` also. `emptyFlag` can’t be boolean?

emro: I completely got rid of this in my update.

src/encoded/static/components/motifs.js
emro: Fixed.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,144 @@
...
+        // Check each element of @graph to see if it has documents and if the available documents are of the type "position weight matrix"
+        // If so, fetch PWM data and insert logo and associated PWM information to a list
+        results.forEach((d) => {
+            if (d.documents[0]) {
+                emptyFlag = 0;
+                if (d.documents[0].document_type === 'position weight matrix') {
+                    pwmLink = urlBase + d.documents[0]['@id'] + d.documents[0].attachment.href;
+
+                    getMotifData(pwmLink, this.context.fetch, entryPoint, d3).then((response) => {
+                        targetElement.insertAdjacentHTML('afterbegin', `<div class="element" id="element${d.accession}"></div>`);
```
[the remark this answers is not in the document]
emro: Avoiding all this `insertAdjacentHTML` now.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,144 @@
...
+        let pwmLink = '';
+        // Check each element of @graph to see if it has documents and if the available documents are of the type "position weight matrix"
+        // If so, fetch PWM data and insert logo and associated PWM information to a list
+        results.forEach((d) => {
+            if (d.documents[0]) {
+                emptyFlag = 0;
+                if (d.documents[0].document_type === 'position weight matrix') {
+                    pwmLink = urlBase + d.documents[0]['@id'] + d.documents[0].attachment.href;
+
+                    getMotifData(pwmLink, this.context.fetch, entryPoint, d3).then((response) => {
```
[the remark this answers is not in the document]
emro: I fixed this.

src/encoded/static/components/motifs.js
emro: I'm not sure if I did exactly the strategy you suggested, but I moved as much code as possible into render() and generally wrote things in a more React way.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,158 @@
...
+            getMotifData(this.pwmLink, this.context.fetch).then((response) => {
+                this.addMotifElement(targetElement, response, entryPoint);
+            });
+        });
+    }
+
+    // Redraw charts when window changes
+    componentDidUpdate() {
+        this.d3 = require('d3');
+        this.sequenceLogos = require('d3-sequence-logo'); // logos This is for local development when changes are needed to d3-sequence-logo.
```
forresttanaka: You already loaded this.d3 and this.sequenceLogos in `componentDidMount`, so you can just use those here. Also I _think_ this `require` might be different from the one in `componentDidMount` because that one came in a parameter where this one gets translated by webpack I’m fairly sure.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,158 @@
...
+                        <p>Biosample: {element.biosample_term_name}</p>
+                    : null}
+                    {element.organ_slims ?
+                        <p>Organ: {element.organ_slims.join(', ')}</p>
+                    : null}
+                    {targetList ?
+                        <p>{targetListLabel}: {targetList.join(', ')}</p>
+                    : null}
+                </div>
+                <div ref={(div) => { this.chartdisplay = div; this.pwmLink = pwmLink; }} className="motif-element" />
```
forresttanaka: The `this.pwmLink` assignment in the ref function is very unusual since it’s not related to the ref, and will break in React 16 since ref functions don’t exist anymore. Since you only use this.pwmLink in a couple places, I think you might as well just calculate `pwmLink` from scratch whenever you need it, or even put the string template into a one-line method so you only need to change it once if it changes.

forresttanaka: …or you could put the assignment in the constructor if you know urlBase and documents won’t change between page re-renders. I don’t know if they can change or not.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,158 @@
...
+    const pwmLinkList = [];
+
+    // Iterate through results to see which have associated PWM data
+    results.forEach((d) => {
+        if (d.documents[0]) {
+            if (d.documents[0].document_type === 'position weight matrix') {
+                pwmLinkList.push(d);
+            }
+        }
+    });
```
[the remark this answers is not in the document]
emro: Using `filter` is a lot more succinct.

src/encoded/static/components/motifs.js
emro: Good point! Fixed.

src/encoded/static/components/motifs.js
emro: I moved this and I think it should work better.

src/encoded/static/components/motifs.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    // Generate PWM link from url and document
+    generatePWMLink() {
+        const element = this.props.element;
+        const urlBase = this.props.urlBase;
+        return `${urlBase}${element.documents[0]['@id']}${element.documents[0].attachment.href}`;
+    }
+
+    addMotifElement(targetElement, response, entryPoint) {
```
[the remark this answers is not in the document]
emro: This is a great point! Updated.

PR #25, opened 2019-05-25
REG-106 fix for biosample in details plus tests

hitz:
Also delete some dead commented code in the testt


hitz:
Hmmm this test *actually* failed.

PR #30, opened 2019-06-14
REG-104 make search result smaller

hitz:
Essentially what this ticket does is exchange "return default dataset fields" with the more usual "columns" definition.  Documents were added to the columns.

### DIFF
--- a/src/encoded/regulome_search.py
+++ b/src/encoded/regulome_search.py
@@ -16,7 +16,8 @@ from encoded.helpers.helper import (
 from snovault.helpers.helper import (
     get_filtered_query,
     set_filters,
-    set_facets
+    set_facets,
+    list_result_fields,
 )
 from .batch_download import get_peak_metadata_links
 from .regulome_atlas import RegulomeAtlas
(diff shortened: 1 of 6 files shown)


src/encoded/static/components/motifs.js
```diff
@@ -125,7 +125,7 @@ export const Motifs = (props) => {
     const urlBase = props.urlBase;
 
     // Filter results to find ones with associated PWM data
-    const pwmLinkList = results.filter(d => d.documents[0] && d.documents[0].document_type === 'position weight matrix');
+    const pwmLinkList = results.filter(d => d.documents && d.documents[0] && d.documents[0].document_type === 'position weight matrix');
```
yunhailuo: Is this right enough? Just look at the first document (which is right for now)?

src/encoded/regulome_search.py
```diff
@@ -533,7 +534,13 @@ def region_search(context, request):
...
         facets = _REGULOME_FACETS
 
-        query = get_filtered_query('Dataset', [], set(), principals, atlas.set_type())
+        query = get_filtered_query(
+            'Dataset',
+            [],
+            sorted(list_result_fields(request, ['Experiment', 'Annotation'])),
+            principals,
+            atlas.set_type()
+        )
```
yunhailuo: Not sure how this work. The third argument and the `list_result_fields` will make elastic return less or trim returns from elastic?

src/encoded/static/components/motifs.js
hitz: I guess this will fail if more than 1 doc is added to these annotation files.   But I don't really want to fabricate a test case for it.   Also there is a (bad) case where there are >1 documents with type = 'position weight matrix' (this shouldn't happen on purpose)   @emro how would you write this to pull all the documents with this type?  My JS is rusty.

src/encoded/regulome_search.py
hitz: list_result_fields will look at the 'columns' for the specified doc_types in the schema.   Empty list (which is what code use to day) defaults to "*".

PR #31, opened 2019-06-18
REG-123 Get feature values from pre-calculated bigWig files

yunhailuo:


src/encoded/regulome_search.py
```diff
@@ -607,12 +611,16 @@ def region_search(context, request):
...
+            evidence = atlas.regulome_evidence(all_hits['datasets'], chromosome, int(start), int(end))
             regdb_score = atlas.regulome_score(all_hits['datasets'], evidence)
+            features = evidence.copy()
+            features.update({
+                k: k in evidence
+                for k in [
+                    'ChIP', 'DNase', 'PWM', 'Footprint', 'eQTL', 'dsQTL',
+                    'PWM_matched', 'Footprint_matched'
+                ]
+            })
```
hitz: This whole bit seems a little clunky.    Why are we copying the evidence Dict then adding back 8 properties?  Seems like it should be one or the other.    Also this is in there twice, should be factored out for the two views.

src/encoded/regulome_atlas.py
```diff
@@ -256,7 +273,7 @@ def _regulome_category(self, score_category=None, dataset=None):
             return 'Single_Nucleotides'
         return '???'
 
-    def regulome_evidence(self, datasets):
+    def regulome_evidence(self, datasets, chrom, start, end):
```
hitz: Should the name of this function stil be regulome_evidence if it's also getting the scores from the bigWig?

yunhailuo: Scores are part of the features (evidence?) being inputted to the regulome scoring algorithm.

src/encoded/regulome_search.py
yunhailuo: `features` is the exact input into the new scoring algorithm. So I removed details and converted values of those binary variables to boolean. Good to have a function to convert `evidence` to `features` @hitz ?

PR #32, opened 2019-06-21
REG-120 index value and strand

yunhailuo:


src/encoded/regulome_indexer.py
```diff
@@ -340,16 +343,29 @@ def region(row):
...
+        if len(row) > 4 and row[3] in ['.', '']:
+            doc['value'] = row[4]
+        return (chrom, doc)
 
     @staticmethod
     def snp(row):
         '''Read a SNP from an in memory row and returns chrom and document to index.'''
         chrom, start, end, rsid = row[0], int(row[1]), int(row[2]), row[3]
+        strand = '.'
+        if len(row) > 5 and row[5] in ['.', '+', '-']:
```
hitz: I think we should write a warning if this condition is not met.

src/encoded/regulome_indexer.py
```diff
@@ -340,16 +343,29 @@ def region(row):
...
+            'coordinates': {
+                'gte': start,
+                'lt': end
+            },
+            'strand': strand,
+        }  # Stored as BED 0-based half open
+        if len(row) > 3:
+            doc['value'] = row[3]
+        if len(row) > 4 and row[3] in ['.', '']:
+            doc['value'] = row[4]
```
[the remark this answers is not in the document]
yunhailuo: Most regulome data will have "info" in column 4, like genes related to eQTL or chromHMM categories. In general, BED column 4 is for name (string). ChIP score will be from column 5 from the standard pipeline.

src/encoded/regulome_indexer.py
yunhailuo: Warning in log?

hitz: good point.  This is in the indexer... I don't think it will write to STDERR/apache log.  Maybe @caseylitton knows.

PR #35, opened 2019-07-16
REG-125 update cloud config from encoded circa v87

caseylitton:


caseylitton:
We're not adding this change as it breaks since we're not up to date with snovault

PR #36, opened 2019-07-17
REG-126-update-to-java-install-manual

caseylitton:


cloud-config.yml
```diff
@@ -193,7 +198,6 @@ runcmd:
 - sudo pip3 install -U zc.buildout setuptools redis pyBigWig
 - sudo -u encoded buildout bootstrap
 - sudo -u encoded LANG=en_US.UTF-8 bin/buildout -c %(ROLE)s.cfg buildout:es-ip=%(ES_IP)s buildout:es-port=%(ES_PORT)s
-- sudo -u encoded bin/aws s3 cp --recursive s3://encoded-conf-prod/.aws .aws
```
hitz: actually this line needs to go back in, because the creds need to be put here as user encoded as well.
Obviously there are other solutions like copying the files to /srv/encoded/.aws as well.

@yunhailuo is going to patch this and try

yunhailuo: Yes. I think putting the line back works.

caseylitton: added it back, resolving convo

PR #37, opened 2019-08-12
REG-124 scoring update

yunhailuo:


src/encoded/regulome_atlas.py
```diff
@@ -365,10 +365,22 @@ def make_a_case(self, snp):
...
+            'dsQTL',
+            'PWM_matched',
+            'Footprint_matched'
+        ]
+        query = [int(k in characterization) for k in binary_keys]
+        numeric_keys = [
+            'IC_max', 'IC_matched_max', 'ChIP_max_signal', 'DNase_max_signal'
+        ]
+        query += [characterization[k] for k in numeric_keys]
+        probability = str(round(TRAINED_REG_MODEL.predict_proba([query])[:, 1][0], 5))
```
hitz: It's a little hard to follow the [:, 1][0], 5)). might be a little clearer if if it was explicit (I assume predict_proba() is returning some kind of 2d array.

But it's not worth re-PRing for this, just a style point.

hitz: actually you may as well fix it since the branch seems to have conflicts.

yunhailuo: I guess I'm not sure what's the explicit way. The model support matrix input where each row is one variant. For us, it's always one variant at a time for the code here. The output is a `<class 'numpy.ndarray'>`. So `[:, 1]` gives me the second column which according to Shengcheng is the probability and `[0]` gives me the number instead of an array. For example,
```python
query = [
    [0, 0, 1, 0, 0, 0, 0, 0, 1.99, 0.0, 1.86, 0.17],
]
TRAINED_REG_MODEL.predict_proba(query)
# array([[0.14, 0.86]])

query = [
    [0, 0, 1, 0, 0, 0, 0, 0, 1.99, 0.0, 1.86, 0.17],
    [1, 1, 0, 0, 0, 0, 0, 0, 0.0, 0.0, 65.24, 0.51],
    [1, 1, 0, 0, 0, 0, 0, 0, 0.0, 0.0, 115.26, 0.63],
    [1, 1, 1, 1, 0, 0, 0, 0, 0.02, 0.0, 267.98, 4.63],
    [1, 1, 0, 0, 0, 0, 0, 0, 0.0, 0.0, 20.92, 0.23],
    [0, 1, 0, 0, 0, 0, 0, 0, 0.0, 0.0, 5.4, 0.25],
    [1, 1, 0, 1, 0, 0, 0, 0, 0.0, 0.0, 634.02, 12.42],
    [1, 1, 1, 1, 1, 0, 0, 0, 1.59, 0.0, 492.81, 7.35],
    [1, 1, 1, 0, 0, 0, 0, 0, 1.71, 0.0, 89.19, 2.73],
]
TRAINED_REG_MODEL.predict_proba(query)
# array([[0.14, 0.86],
#        [0.42, 0.58],
#        [0.35, 0.65],
#        [0.39, 0.61],
#        [0.12, 0.88],
#        [0.37, 0.63],
#        [0.29, 0.71],
#        [0.25, 0.75],
#        [0.31, 0.69]])
```

Any suggestions for making it explicit, @hitz ?

hitz: maybe just include the expected format as a comment and say you are getting the first value of the second column or whatever.

yunhailuo: Comments added

PR #38, opened 2019-08-14
REG-74 Summary graphic and clickable thumbnails

emro:
There are several big changes here:
(1) thumbnails with mini charts/images are now used to filter the data and provide different views
(2) there is a new clickable/expandable vertical bar chart for "Accessibility" data
(3) a new component draws and svg diagram of the searched-for SNP and the nearby SNPs


src/encoded/static/components/visualizations.js
```diff
@@ -244,13 +246,273 @@ export class TestViz extends React.Component {
...
+    handleClick(clickID) {
+        if (!(this.state.currentTarget.includes(`table${clickID}`))) {
+            this.setState({
+                currentTarget: [...this.state.currentTarget, `table${clickID}`],
+            });
+        } else {
+            const removedTarget = this.state.currentTarget.filter(element => element !== `table${clickID}`);
+            this.setState({
+                currentTarget: removedTarget,
+            });
```
forresttanaka: The cart code does something like this to remove an element. I don’t think it’s guaranteed to perform faster than this, but filter does have to go through the entire array, and I don’t think this does. If currentTarget can’t get very long, then this probably doesn’t matter:

```
            const doomedIndex = state.elements.indexOf(action.elementAtId);
            if (doomedIndex !== -1) {
                return Object.assign({}, state, {
                    elements: state.elements
                        .slice(0, doomedIndex)
                        .concat(state.elements.slice(doomedIndex + 1)),
                });
            }
```

src/encoded/static/components/visualizations.js
```diff
@@ -244,13 +246,273 @@ export class TestViz extends React.Component {
...
+            sortedFakeFacets[key] = fakeFacets[key];
+        });
+        const chartData = sortedFakeFacets;
+        const chartTitle = this.props.displayTitle;
+        const chartArray = Object.keys(chartData).map(key => chartData[key]);
+        const chartMax = Math.max(...chartArray);
+        // compute left margin
+        let leftMargin = 60;
+        Object.keys(chartData).forEach((d) => {
+            leftMargin = Math.max(d.length * 7, leftMargin);
```
forresttanaka: This looks like it just gets the last chartData array length or leftMargin.

src/encoded/static/components/visualizations.js
```diff
@@ -244,13 +246,273 @@ export class TestViz extends React.Component {
...
+            leftMargin,
+            data,
+        });
+    }
+
+    render() {
+        let errorMessage = false;
+        if (this.state.searchTerm) {
+            errorMessage = true;
+        }
```
forresttanaka: This looks the same as `const errorMessage = !!this.state.searchTerm`.

src/encoded/static/components/visualizations.js
```diff
@@ -244,13 +246,273 @@ export class TestViz extends React.Component {
...
+                        searchTermMatch = sanitizedString(d).match(this.state.searchTerm);
+                        if (searchTermMatch) {
+                            errorMessage = false;
+                        }
+                    }
+                    return (
+                        <div
+                            className={`biosample-table table${dIDX} ${searchTermMatch ? 'display-table' : ''}`}
+                            key={`table${dIDX}`}
+                        >
```
forresttanaka: If you can avoid using a loop index as a key (and I’m not sure if you can here or not) and instead use some data unique to each array element, that could avoid problems if the order of chartData changes.

forresttanaka (changes requested):
I didn’t look over your SCSS, but figured you might need this sooner than later.

src/encoded/static/components/motifs.js
```diff
@@ -63,10 +63,7 @@ export class MotifElement extends React.Component {
 
     // Redraw charts when window changes
     componentDidUpdate() {
-        // Need to remove current logo or another logo will be appended to original
-        this.chartdisplay.innerHTML = '';
         const pwmLink = this.generatePWMLink();
```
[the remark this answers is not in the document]
emro: I moved that into the PWM library.

src/encoded/static/components/visualizations.js
emro: I think I need this loop because I have to find the biosample name with the most characters and that determines what the left margin should be.

src/encoded/static/components/regulome_search.js
```diff
@@ -255,67 +244,296 @@ PeakDetails.propTypes = {
...
+                        } else {
+                            offset = 20;
+                        }
+                    } else {
+                        offset = 20;
+                    }
+                }
+            }
+            coordinateYOffset.push(offset);
+        });
```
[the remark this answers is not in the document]
emro: I don't think I know how to do this with a `.map`. As I go through the nearby SNPs, I have to check the previous values that I have assigned to 'coodinateYOffset` in order to assign the current one. Can I do that with `.map`?

src/encoded/static/components/regulome_search.js
```diff
@@ -362,89 +577,134 @@ class RegulomeSearch extends React.Component {
...
-                                }
-                                {(context.regulome_score && !context.peak_details) ?
-                                    <a
-                                        rel="nofollow"
-                                        className="peaks-link"
-                                        href={searchBase ? `${searchBase}&peak_metadata` : '?peak_metadata'}
+                                        </div>
+                                    : null}
+                                </button>
+                                <button
```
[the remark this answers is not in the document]
emro: Well, I'm not sure the best way to deal with this. I could make the buttons not html buttons, just clickable divs but it seems better to have them as buttons. But there most definitely must be divs in the thumbnails.

src/encoded/static/components/visualizations.js
forresttanaka: I see that now. nvm.

src/encoded/static/components/regulome_search.js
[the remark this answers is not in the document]
forresttanaka: I see. No, I don’t think you can use .map() then.

src/encoded/static/components/visualizations.js
emro: It shouldn't get too long but I think slice() is a better choice anyway.

forresttanaka: `removedTarget` needs to be a memory copy of this.state.currentTarget to use with setState, where here you modify this.state (https://stackoverflow.com/questions/26253351/correct-modification-of-state-arrays-in-reactjs/#26254086). You’ll need to pass a complete copy of this.state.currentTarget with an element removed.

forresttanaka (changes requested):
Looks good! Just one change needed for a setState.

forresttanaka (approved):
Looked over the changes, and I think they look good!

forresttanaka (approved):
Just one note that I really don’t know if it’s even better or not.

src/encoded/static/components/regulome_search.js
```diff
@@ -575,6 +569,12 @@ class RegulomeSearch extends React.Component {
         const context = this.props.context;
         const notification = context.notification;
         const urlBase = this.context.location_href.split('/regulome-search')[0];
+        const allData = context['@graph'];
+        const QTLData = allData.filter(d => (d.annotation_type && d.annotation_type.indexOf('QTL') !== -1));
+        const chipData = allData.filter(d => d.assay_title === 'ChIP-seq');
+        const dnaseData = allData.filter(d => (d.assay_title === 'FAIRE-seq' || d.assay_title === 'DNase-seq'));
+        const chromatinData = allData.filter(d => (d.annotation_type === 'chromatin state'));
```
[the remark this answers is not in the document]
emro: It doesn't make the code shorter, but it does seem neater!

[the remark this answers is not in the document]
emro: I'm actually going to revert this because when groupedData is {}, then the subsequent lines cause errors and .filter doesn't have that problem.

PR #39, opened 2019-08-16
REG-130 fix query

yunhailuo:


src/encoded/regulome_search.py
```diff
@@ -296,10 +296,9 @@ def get_coordinate(query_term, assembly='GRCh37', atlas=None):
 
 
 def get_rsids(atlas, assembly, chrom, start, end):
-    pos = (int(start) + int(end)) / 2
-    window = int(end) - int(start)
-    rsids = atlas.nearby_snps(_GENOME_TO_ALIAS.get(assembly, 'hg19'),
-                              chrom, pos, window=window, max_snps=99999)
+    rsids = atlas.find_snps(
```
hitz: Why do you suppose get_rsids was using nearby_snps() in the first place...

hitz: maybe the old version of find_snps didn't take a range.

yunhailuo: Could be. I couldn't remember. But I think `find_snps` here is more proper for now.

hitz:
This has conflicts again, probably with 74

yunhailuo:
> This has conflicts again, probably with 74

Right. Working on it now.

PR #40, opened 2019-08-22
REG-128 use peak metadata

yunhailuo:


src/encoded/static/components/motifs.js
```diff
@@ -85,22 +85,22 @@ export class MotifElement extends React.Component {
 
     render() {
         const element = this.props.element;
-        const targetList = element.targets.map(t => t.label);
+        const targetList = element.targets;
         const targetListLabel = targetList.length > 1 ? 'Targets' : 'Target';
 
         return (
-            <div className="element" id={`element${element.accession}`}>
+            <div className="element" id={`element${element.dataset.split('/')[2]}`}>
```
emro: I would probably put `const accession = element.dataset.split('/')[2]` above `return(` and then use that on lines 92 and 95.

yunhailuo: Sure. Will do.

yunhailuo: Fixed.

PR #41, opened 2019-09-05
REG-91 chromatin visualization

emro:
Note: two of the commits are from @yunhailuo's PR. This branch should be merged after his PR, and the changes from those commits do not need to be reviewed here.

This branch is for chromatin visualizations: it adds a bar chart on the chromatin thumbnail and an interactive table on the chromatin tab, where the table results are controlled by a search bar (searchable by organ, biosample, or chromatin state) and by a bar chart showing how many results correspond to each chromatin state. Unlike the other visualizations, the bar charts for chromatin are ordered by state, not by number of results.

### DIFF
--- a/src/encoded/static/components/datacolors.js
+++ b/src/encoded/static/components/datacolors.js
@@ -1,3 +1,6 @@
+import _ from 'underscore';
+
+
 // DataColors is a class for managing symbolic data colors throughout the app. Essentially it
 // statically maps data keys to colors consistently even when the data changes.
 //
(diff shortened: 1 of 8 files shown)


src/encoded/static/components/datacolors.js
```diff
@@ -1,3 +1,6 @@
+import _ from 'underscore';
+
```
emro: All of these changes are directly from Encode: I copied the file from Encode over to Regulome.

src/encoded/static/components/visualizations.js
```diff
@@ -496,6 +606,335 @@ export class ChartTable extends React.Component {
...
+ChartList.propTypes = {
+    data: PropTypes.array.isRequired,
+    displayTitle: PropTypes.string.isRequired,
+    chartWidth: PropTypes.number.isRequired,
+    dataFilter: PropTypes.string.isRequired,
+};
+
+// Display heat map for chromatin states
+// Note: this was developed and then it was decided that this was the wrong way to visualize the data
+// Nonetheless it may be useful later
```
emro: This will probably not be useful in the future but I was considering leaving it just in case.

src/encoded/static/components/visualizations.js
```diff
@@ -1,5 +1,63 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { ResultsTable } from './regulome_search';
```
emro: It is probably not ideal to be importing components from "regulome_search" into "visualizations" and from "visualizations" into "regulome_search" but I was having trouble thinking of a straightforward way to get around it.

src/encoded/static/components/visualizations.js
```diff
@@ -496,6 +606,335 @@ export class ChartTable extends React.Component {
...
+        const chartKeys = Object.keys(chartData);
+        const chartArray = chartKeys.map(key => chartData[key]);
+        const chartMax = Math.max(...chartArray);
+        let selectedStates = [];
+        for (let idx = 0; idx < chartKeys.length; idx += 1) {
+            if (chartData[chartKeys[idx]] > 0) {
+                selectedStates = [sanitizedString(chartKeys[idx])];
+                break;
+            }
+        }
```
forresttanaka: This looks like a chartKeys.find()? Then sanitize the result and put it in a single-element array? Or is that not the same?

emro: I think we need to leave it like this. Here we are initializing the selected state to be most active state for which there are results. Often that will be "Active TSS" but not always. Since we don't know what the selected state should be, we can't use `.find`.

hitz:
This conflicts because it needs a rebase

src/encoded/static/components/visualizations.js
```diff
@@ -463,21 +567,27 @@ export class ChartTable extends React.Component {
...
-                                {this.state.data.filter(element => filterForBiosample(element, d)).map(d2 =>
-                                    <div className="table-entry" key={`table-entry-${d2.accession}`}>
-                                        <p><a href={d2['@id']}>{d2.accession}</a></p>
+                                {this.state.data.filter(element => filterForKey(element, d, this.props.dataFilter)).map(d2 =>
+                                    <div className="table-entry" key={`table-entry-${d2.dataset.split('/')[2]}`}>
+                                        <p><a href={d2.dataset}>{d2.dataset.split('/')[2]}</a></p>
                                         <div className="inset-table-entries">
-                                            {d2.organ_slims ?
-                                                <p><span className="table-label">Organ</span>{d2.organ_slims.join(', ')}</p>
+                                            {d2.biosample_ontology.organ_slims ?
```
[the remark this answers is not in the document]
emro: I'm not sure, but we can just check.

src/encoded/static/components/visualizations.js
```diff
@@ -1,5 +1,63 @@
...
+    : name.includes('Enh') ? 'Enhancers'
+    : name.includes('ZNF/Rpts') ? 'ZNF genes & repeats'
+    : name.includes('Het') ? 'Heterochromatin'
+    : name.includes('TssBiv') ? 'Bivalent/Poised TSS'
+    : name.includes('BivFlnk') ? 'Flanking Bivalent TSS/Enh'
+    : name.includes('ReprPCWk') ? 'Weak Repressed PolyComb'
+    : name.includes('ReprPC') ? 'Repressed PolyComb'
+    : name.includes('Quies') ? 'Quiescent/Low'
+    : null
+);
```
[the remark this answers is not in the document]
emro: I've changed this and think it should be good, but some of the state abbreviations that we are using may not match the ones on my reference and then I may need to update again. For instance, we have "1_TssA" instead of "TssA" and I got around this by using 'includes'

src/encoded/static/components/visualizations.js
emro: I think I will not worry about this one for now, maybe if there are more components like this in the future.

hitz:
Confirmed that non-bdd tests pass, at least sometimes.

PR #42, opened 2019-09-10
REG-116 Genome browser

emro:
This branch adds the genome browser thumbnail and tab. There are multiple datasets associated with each SNP, so this means we need to do queries in order to retrieve all the visualizable files that correspond to the datasets. We don't want to display every visualizable file, however, just 2 per dataset, so we have specific queries and filtering in order to limit the results that are retrieved and displayed. Here is how we are choosing files:

DNase → peaks, read-depth normalized signal (rep1)
ChIP → peaks and background as input for IDR, signal p-value (rep1,2) or rep1
FAIRE → peaks, signal

The other feature that is different from Encode is the pagination: we do not display all the results in one long list, instead results are split up across pages. Ultimately we will also have filters so you can search for a biosample or target.


src/encoded/static/components/genome_browser.js
```diff
@@ -1,610 +1,501 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import _ from 'underscore';
+import { FetchedData, Param } from './fetched';
+import { BrowserFeat } from './browserfeat';
```
emro: This file was copied over from Encode. There have been some small changes but not too many.

src/encoded/static/components/genome_browser.js
```diff
@@ -1,610 +1,501 @@
...
+            visualizer: null,
+            showAutoSuggest: true,
+            searchTerm: '',
+            genome: '',
+            contig: props.coordinates.split(':')[0],
+            x0: +props.coordinates.split(':')[1].split('-')[0] - 5000,
+            x1: +props.coordinates.split(':')[1].split('-')[1] + 5000,
+            pinnedFiles: [],
+            disableBrowserForIE: false,
+            geneSearch: false,
```
emro: I may need a better solution for this. We don't want the gene search right now because the back end doesn't support it, but we want to add it back later. I don't want to remove all the code that I wrote to support autocomplete, so this is a way to turn that feature off for now.

hitz:
THis conflicts because it needs rebase

src/encoded/static/components/regulome_search.js
```diff
@@ -690,8 +912,11 @@ class RegulomeSearch extends React.Component {
                                 >
                                     <h4>QTL Data</h4>
                                     {(this.state.selectedThumbnail === null) ?
-                                        <div className="line"><i className="icon icon-chevron-circle-right" />Click to see dsQTL and eQTL data.
-                                            <div>(<b>{QTLData.length}</b> result{QTLData.length !== 1 ? 's' : ''})</div>
+                                        <div>
+                                            <div className="line"><i className="icon icon-chevron-circle-right" />Click to see dsQTL and eQTL data.
+                                                <div>(<b>{QTLData.length}</b> result{QTLData.length !== 1 ? 's' : ''})</div>
+                                            </div>
+                                            <QTLDetails {...this.props} />
```
forresttanaka: Looks like `QTLDetails` only takes one property: context. Sometimes passing a spread makes it tough to hunt down what the called component actually requires.

forresttanaka (changes requested):
Like the last one, I don’t think any of my comments is required, so let me know if you want to make no changes and I’ll approve it. The one big suggestion is your big GET request(s), but I’m not sure if my suggestion is practical.

forresttanaka (approved):

hitz:
Needs rebase

emro:
I rebased and some updates were required because of the new data structure but I think it's all set now.

forresttanaka (approved):
I mostly reviewed the past three commits. My guess on the duplicate datasets is either a data issue, or a misunderstanding of some kind. I feel uncomfortable deleting the duplicates, but there have been cases of things in encoded I was uncomfortable with, only to turn out to be exactly what wranglers/labs wanted. My intuition runs about 50/50 on correctness, if I’m generous.

PR #45, opened 2019-09-18
REG-121 update QTL thumbnail

emro:
The QTL thumbnail summarizes the full table with 3 columns:
method (ds or e QTL), biosample, and target (eQTL) 
Target for dsQTLs are always blank, so the target should map to N/A.

### DIFF
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -77,7 +77,7 @@ const dataColumnsQTL = {
     },
     value: {
         title: 'Targets',
-        getValue: item => (item.value ? item.value : 'N/A'),
+        getValue: item => item.value || 'N/A',
     },
     dataset: {
         title: 'Dataset',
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/regulome_search.js
```diff
@@ -77,6 +77,7 @@ const dataColumnsQTL = {
     },
     value: {
         title: 'Targets',
+        getValue: item => (item.value ? item.value : 'N/A'),
```
forresttanaka: `item.value || 'N/A'`? Similar below for dataColumnsQTLShort.value.

src/encoded/static/components/regulome_search.js
```diff
@@ -478,6 +476,8 @@ export const ResultsTable = (props) => {
     let dataColumns = null;
     if (props.dataFilter === 'chromatin') {
         dataColumns = dataColumnsChromatin;
+    } else if (props.dataFilter === 'qtl' && props.shortened) {
```
forresttanaka: _Seems_ like you could just put the `props.shortened` test inside the test against 'qtl' to avoid the second `props.dataFilter === 'qtl'`.

forresttanaka (approved):
I left a couple suggestions, but I don’t think either of them makes much of a difference, performance or maintenance wise.

forresttanaka (approved):

PR #47, opened 2019-09-23
REG-143 fix browser issues

emro:
This ticket fixes a number of minor style-related issues, among them:
- old versions of Safari do not display rgba correctly
- fonts are generally smaller on mobile
- thumbnail labels stack better on mobile
- number of bars on mobile bar chart is reduced
- chromatin visualization is rearranged to be more readable
- search bar text is shortened so it doesn't overflow
- ipad has 2x3 grid instead of 1x6 grid
- hamburger menu is functional / actually shows up on mobile

### DIFF
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -9,6 +9,9 @@ import { BarChart, ChartList, ChartTable, lookupChromatinNames } from './visuali
 import { requestSearch } from './objectutils';
 import GenomeBrowser from './genome_browser';
 
+const screenMediumMax = 787;
+const screenSmallMax = 483;
+
 const dataTypeStrings = [
     {
         type: 'dbSNP IDs',
@@ -607,10 +610,13 @@ class RegulomeSearch extends React.Component {
         if (this.applicationRef) {
             const screenWidth = this.applicationRef.offsetWidth;
             let thumbnailWidth = 0;
-            if (screenWidth > 787) {
+            // for desktop screens, the display is 3x2 thumbnails
+            if (screenWidth > screenMediumMax) {
                 thumbnailWidth = (this.applicationRef.offsetWidth / 3) - 40;
-            } else if (screenWidth > 483) {
+            // for vertical tablets, the display is 2x3 thumbnails
+            } else if (screenWidth > screenSmallMax) {
                 thumbnailWidth = (this.applicationRef.offsetWidth / 2) - 40;
+            // anything narrower than a vertical tablet has a 1x3 display
             } else {
                 thumbnailWidth = this.applicationRef.offsetWidth - 40;
             }
@@ -624,9 +630,9 @@ class RegulomeSearch extends React.Component {
     addNewFiles(searchQuery) {
         return new Promise((ok) => {
             requestSearch(searchQuery).then((results) => {
-                const newFiles = results ? results['@graph'] : [];
+                const newFiles = (results && results['@graph']) ? results['@graph'] : [];
                 // only update state if new files have been retrieved
-                if (newFiles) {
+                if (newFiles.length > 0) {
                     this.setState(prevState => ({
                         allFiles: [...prevState.allFiles, ...newFiles],
                     }));
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -622,9 +622,12 @@ class RegulomeSearch extends React.Component {
         return new Promise((ok) => {
             requestSearch(searchQuery).then((results) => {
                 const newFiles = results ? results['@graph'] : [];
-                this.setState(prevState => ({
-                    allFiles: [...prevState.allFiles, ...newFiles],
-                }));
+                // only update state if new files have been retrieved
+                if (newFiles) {
+                    this.setState(prevState => ({
+                        allFiles: [...prevState.allFiles, ...newFiles],
+                    }));
+                }
                 ok('success');
             });
         });
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -605,8 +605,10 @@ class RegulomeSearch extends React.Component {
     updateDimensions() {
         const screenWidth = this.applicationRef.offsetWidth;
         let thumbnailWidth = 0;
-        if (screenWidth > 865) {
+        if (screenWidth > 787) {
             thumbnailWidth = (this.applicationRef.offsetWidth / 3) - 40;
+        } else if (screenWidth > 483) {
+            thumbnailWidth = (this.applicationRef.offsetWidth / 2) - 40;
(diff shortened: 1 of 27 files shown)


src/encoded/static/components/visualizations.js
```diff
@@ -524,18 +531,28 @@ export class ChartList extends React.Component {
...
+                    let leftMargin = 0;
+                    if (chartWidth > 500) {
+                        leftMargin = this.state.leftMargin;
+                        barWidth = ((chartWidth - leftMargin) / this.state.chartMax) * this.state.chartData[d];
+                        remainderWidth = chartWidth - barWidth - leftMargin;
+                    } else {
+                        leftMargin = '100%';
+                        barWidth = (chartWidth / this.state.chartMax) * this.state.chartData[d];
+                        remainderWidth = chartWidth - barWidth;
+                    }
```
forresttanaka: Takes some time for me to figure out what this if/else is for.

src/encoded/static/scss/encoded/modules/_regulome_search.scss
```diff
@@ -734,21 +774,28 @@ table {
...
         max-width: 32%;
         align-items: center;
         padding: 10px 15px;
-        transition: flex 0.3s ease-in-out;
-        -moz-transition: flex 0.3s ease-in-out;
-        -webkit-transition: flex 0.3s ease-in-out;
+        transition: flex 0.3s ease-in-out, font-size 0.3s ease-in-out;
+        -moz-transition: flex 0.3s ease-in-out, font-size 0.3s ease-in-out;
+        -webkit-transition: flex 0.3s ease-in-out, font-size 0.3s ease-in-out;
         color: #333;
```
forresttanaka: Doesn’t _seem_ like you need the browser prefixes for transition (https://caniuse.com/#feat=css-transitions), or did you find a situation where you needed these?

src/encoded/static/scss/encoded/modules/_regulome_search.scss
```diff
@@ -1143,6 +1274,21 @@ button {
...
+        @media screen and (max-width: $screen-sm-min) {
+            height: 28px;
+        }
+        .bar {
+            height: 24px;
+            @media screen and (max-width: $screen-sm-min) {
+                height: 28px;
+            }
+        }
+    }
```
forresttanaka: Interesting you always use max-width (desktop first) and I always use min-width (mobile first). At least in encoded, we might need to settle on something.

src/encoded/static/components/visualizations.js
emro: Should I add a comment? I didn't because it was in render() but I could clarify somewhere.

src/encoded/static/scss/encoded/modules/_regulome_search.scss
emro: No, just a habit.

src/encoded/static/components/visualizations.js
forresttanaka: It’s up to you, but it just looked difficult for someone else to maintain it later. It’s in a render() return, but it’s outside of JSX, so comments aren’t awkward, in my view. Comments in JSX are kind of awkward, so I haven’t done that very much.

src/encoded/static/scss/encoded/modules/_regulome_search.scss
emro: OK! We should probably stick with min-width for Encoded since that's already the convention.

emro:
It looks like I need to rebase this.

src/encoded/static/components/regulome_search.js
```diff
@@ -593,26 +593,34 @@ class RegulomeSearch extends React.Component {
...
 
     addNewFiles(searchQuery) {
         return new Promise((ok) => {
             requestSearch(searchQuery).then((results) => {
                 const newFiles = results ? results['@graph'] : [];
-                this.setState(prevState => ({
-                    allFiles: [...prevState.allFiles, ...newFiles],
-                }));
+                // only update state if new files have been retrieved
+                if (newFiles) {
```
forresttanaka: `newFiles` would always be truthy. If you meant `newfiles.length > 0` that would also take care of the case where `results['@graph'].length` is zero.

src/encoded/static/components/regulome_search.js
```diff
@@ -593,26 +593,34 @@ class RegulomeSearch extends React.Component {
...
-            thumbnailWidth = this.applicationRef.offsetWidth - 40;
+        // check for applicationRef because otherwise there are errors during resizing process
+        if (this.applicationRef) {
+            const screenWidth = this.applicationRef.offsetWidth;
+            let thumbnailWidth = 0;
+            if (screenWidth > 787) {
+                thumbnailWidth = (this.applicationRef.offsetWidth / 3) - 40;
+            } else if (screenWidth > 483) {
+                thumbnailWidth = (this.applicationRef.offsetWidth / 2) - 40;
+            } else {
```
forresttanaka: By embedding these numbers like "787" and "483" in the code, could this cause a broader problem later if multiple places in the code need the same breakpoints, or if someone else needs to know what these widths are for? Maybe we need something sort of like the $screen-sm-min constants, but in JS.

src/encoded/static/components/regulome_search.js
```diff
@@ -614,9 +620,9 @@ class RegulomeSearch extends React.Component {
     addNewFiles(searchQuery) {
         return new Promise((ok) => {
             requestSearch(searchQuery).then((results) => {
-                const newFiles = results ? results['@graph'] : [];
+                const newFiles = (results && results['@graph']) ? results['@graph'] : [];
```
forresttanaka: "@graph" is guaranteed to be there though its length might be zero, so the way you had it before should be fine. The important part was checking `newFiles`’s length instead of its existence.

emro: Well, locally I was getting errors without the `&& results['@graph']` because `results['@graph']` would be undefined. This way I see the genome and genes tracks when I run it locally.

forresttanaka: Oh! Maybe regulome is different. OK.

src/encoded/static/components/summary.js
```diff
@@ -297,6 +297,7 @@ class SummaryVerticalFacets extends React.Component {
 
         return (
             <div className="summary-content__facets-vertical">
+                <FilterList {...this.props} />
```
forresttanaka: This is purely a personal preference (and I might have mentioned it before anyway) but I prefer to use destructuring as properties to a component as rarely as possible, because I sometimes wonder where certain properties came from, and destructuring can obfuscate that. FilterList only takes one property anyway, so it would seem easy to just use that explicitly.

emro: OK, I think it is good now.

PR #51, opened 2019-10-07
REG-147 upgrade React

emro:
Upgrading React 15 to React 16.


src/encoded/static/components/regulome_search.js
```diff
@@ -846,14 +846,16 @@ class RegulomeSearch extends React.Component {
                                 >
                                     <h4>ChIP data</h4>
                                     {(thumbnail === null) ?
-                                        <div>
+                                        <React.Fragment>
                                             <div className="line"><i className="icon icon-chevron-circle-right" />Click to see detailed ChIP-seq data.
-                                                <div>(<b>{chipData.length}</b> result{chipData.length !== 1 ? 's' : ''})</div>
+                                                <React.Fragment>
```
forresttanaka: Does this make the “result” line move inline with “Click to see detailed ChIP-seq data”? Without the `<div>`, that’s what this looks like, unless `<React.Fragment>` inserts something automatically in this case.

src/encoded/static/components/search.js
```diff
@@ -1451,7 +1452,7 @@ export class ResultTable extends React.Component {
         }
 
         return (
-            <div>
+            <React.Fragment>
```
forresttanaka: Since only one `<div>` is inside, could you just delete the previous `<div>` and replace it with nothing?

src/encoded/static/components/search.js
```diff
@@ -1625,9 +1626,9 @@ export const ResultBrowser = (props) => {
         );
     }
     return (
-        <div>
+        <React.Fragment>
```
forresttanaka: Similarly, seems like you could just delete the `<div>` as I could have done in encoded but missed this.

src/encoded/static/components/auditmatrix.js
```diff
@@ -228,7 +228,7 @@ class AuditMatrix extends React.Component {
             const type = pluralize(queryStringType.toLocaleLowerCase());
 
             return (
-                <div>
+                <React.Fragment>
```
forresttanaka: Could the previous `<div>` have just been deleted? It only surrounds one div. Not sure if there was a CSS style that relied on it or not.

src/encoded/static/components/regulome_summary.js
```diff
@@ -48,17 +48,17 @@ const snpsColumns = {
 const SNPSummary = (props) => {
     const snps = props.context.summaries;
     return (
-        <div>
+        <React.Fragment>
```
forresttanaka: Doesn’t look like the old `<div>` was necessary in the first place. I think `<SortTablePanel>` could be the outer component here.

src/encoded/static/components/regulome_search.js
emro: Yes, it does move it inline. I needed this <div>! (And the others like it.)

PR #53, opened 2019-10-08
REG-154 remove ChIP and DNase signals from scoring algorithm

yunhailuo:


src/encoded/regulome_atlas.py
```diff
@@ -250,10 +244,8 @@ def _score_category(dataset):
             return 'PWM'
         if collection_type == 'Footprints':
             return 'Footprint'
-        if collection_type in ['eQTLs', 'curated SNVs']:
-            return 'eQTL'
-        if collection_type == 'dsQTLs':
-            return 'dsQTL'
+        if collection_type in ['eQTLs', 'dsQTLs', 'curated SNVs']:
```
hitz: I don't understand this change.

PR #57, opened 2019-10-25
REG-137 add filters to browser

emro:
Creates facets for the genome browser:
- File type
- Organ (typeahead)
- Biosample (typeahead)
- Method
- Target (typeahead)

There are also some very minor style changes:
- I added a background on the searched-for SNP on the line diagram
- I fixed the width of the SNP annotation background rectangles to be more accurate on the line diagram
- I moved the pagination buttons and "Page X of Y" to be stacked

### DIFF
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -6,11 +6,19 @@ import * as globals from './globals';
 import { SortTablePanel, SortTable } from './sorttable';
 import { Motifs } from './motifs';
 import { BarChart, ChartList, ChartTable, lookupChromatinNames } from './visualizations';
-import { requestSearch } from './objectutils';
+import { requestSearch, shadeOverflowOnScroll } from './objectutils';
 import GenomeBrowser from './genome_browser';
 
 const screenMediumMax = 787;
 const screenSmallMax = 483;
+
+// Number of terms to show, the rest will be viewable on scroll
+const displayedTermsCount = 5;
+
+// Define facets (probably this should be in a schema really)
+const facetList = ['file_format', 'organ', 'biosample', 'assay', 'target'];
+const facetTitleList = ['File format', 'Organ / cell type', 'Biosample', 'Method', 'Target'];
+const typeaheadFacetList = [false, true, true, false, true];
 
 const dataTypeStrings = [
     {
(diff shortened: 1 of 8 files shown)


src/encoded/static/components/regulome_search.js
```diff
@@ -694,10 +1063,12 @@ class RegulomeSearch extends React.Component {
             const biosampleMap = {};
             const assayMap = {};
             const targetMap = {};
+            const organMap = {};
             experimentDatasets.forEach((dataset) => {
                 biosampleMap[dataset.dataset] = dataset.biosample_ontology.term_name || '';
                 assayMap[dataset.dataset] = dataset.method || '';
                 targetMap[dataset.dataset] = dataset.targets ? dataset.targets.join(', ') : '';
+                organMap[dataset.dataset] = (dataset.biosample_ontology.classification === 'tissue') ? dataset.biosample_ontology.organ_slims.join(', ') : dataset.biosample_ontology.cell_slims.join(', ');
```
emro: This is the line where we sometimes add `organ_slims` and sometimes add `cell_slims`.

emro:
One thing I was debating right now is if it would make the more sense to make one component that generates the filter buttons and the facet buttons, since they both rely on the same `addGenomeFilter` function. Do you have thoughts about that?

forresttanaka:
`FacetButton` and `FilterButton` look reasonably different to me, so I don’t see a lot of benefit combining those. They’re both small, and both rely on a shared method, so it looks like reasonable design to me.

src/encoded/static/components/regulome_search.js
```diff
@@ -538,27 +538,29 @@ ResultsTable.defaultProps = {
...
+
+    render() {
+        const { buttonLabel, lookupFilterCount, selectedFacets, facetLabel } = this.props;
+        return (
+            <button
+                className={selectedFacets.includes(`${buttonLabel}AND${facetLabel}`) ? 'active' : ''}
+                onClick={this._addGenomeFilter}
+                disabled={(lookupFilterCount(buttonLabel, facetLabel) === ' (0)')}
+            >
+                {buttonLabel}{lookupFilterCount(buttonLabel, facetLabel)}
```
forresttanaka: The `lookupFilterCount` implementation doesn’t look entirely trivial, and it would get called with the same parameters twice in a row here. Maybe just call it once?

emro:
A couple more changes from QA feedback:
- We're changing the colors for the base pairs to match a more conventional color scheme: A=green, G=yellow, T=red, C=blue
- We are sorting the facet terms so that ones that are clickable or selected are displayed above the ones that are disabled.
- We are displaying a number in parentheses next to the selected terms which indicates how many of the results match that term.

src/encoded/static/components/regulome_search.js
```diff
@@ -759,17 +781,44 @@ class GenomeFacets extends React.Component {
...
+        return arr.filter(el => this.lookupFilterCount(el, category) !== ' (0)');
+    }
+
+    isZero(arr, category) {
+        return arr.filter(el => this.lookupFilterCount(el, category) === ' (0)');
+    }
+
+    placeZerosAtEnd(arr, facet) {
+        let arrBegin = [...arr];
+        arrBegin = this.isntZero(arrBegin, facet).sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
```
forresttanaka: It looks to me like you could pass `arr` to `this.isntZero` instead of making a copy in `arrBegin` first because `this.isntZero` makes a filtered copy of its first parameter anyway, and doesn’t mutate anything. Same thing for `arrEnd` below.

src/encoded/static/components/regulome_search.js
```diff
@@ -662,6 +668,19 @@ const filterByAllSelectedFilters = (files, facets) => {
     return newFiles;
 };
 
+const filterByOneFilter = (files, facet) => {
+    const facetFilter = facet.split('AND')[0];
+    const facetName = facet.split('AND')[1];
+    let newFiles = files;
```
forresttanaka: You’re not making a copy of `files` here, but it seems like it’d be easy to understand using `files.filter` below, and just using `let newFiles;` here.

PR #62, opened 2019-11-21
REG-164 update motif colors

emro:
We want all the base pair colors to match (on the logos and on the genome browser):
A = green
G = yellow
T = red
C = blue

### DIFF
--- a/src/encoded/static/libs/d3-sequence-logo.js
+++ b/src/encoded/static/libs/d3-sequence-logo.js
@@ -230,10 +230,10 @@ function entryPoint(logoSelector, PWM, d3, alignmentCoordinate, firstCoordinate,
     const svgLetterHeight = 150;
 
     const colorBase = {
-        A: '#C13B42',
-        C: '#EFB549',
-        G: '#335C95',
-        T: '#489655',
+        A: '#489655',
+        C: '#335C95',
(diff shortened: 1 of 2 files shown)


src/encoded/static/libs/d3-sequence-logo.js
```diff
@@ -230,10 +230,10 @@ function entryPoint(logoSelector, PWM, d3, alignmentCoordinate, firstCoordinate,
     const svgLetterHeight = 150;
 
     const colorBase = {
-        A: '#C13B42',
```
zoldello: This change looks fine, it is just a color change. I am curious, how were the colors picked.

emro: Green, yellow, red, and blue are the standard colors for the base pairs. It seems like the convention of which color corresponds to which base pair is not very well defined and different people use different conventions. Originally, the code had `const colors = d3.scaleOrdinal(d3.schemeCategory10);` to define the colors, but I updated that by picking particular tones of green, yellow, red, and blue. However, we needed to change the colors again because our colors did the match the convention that Alan from the University of Michigan wanted us to use, and additionally we needed them to match what we are using for the base pairs on the genome browser.

PR #63, opened 2019-11-22
REG-168 fix ENCODE file download

yunhailuo:


src/encoded/types/file.py
```diff
@@ -555,6 +555,11 @@ def download(context, request):
     mapping = context.schema['file_format_file_extension']
     file_extension = mapping[properties['file_format']]
     accession_or_external = properties.get('accession') or properties['external_accession']
+    if accession_or_external.startswith('ENCFF'):
```
hitz: man that is a FILTHY hack.  But real solution is hard.    Can you put some comments in the code?

yunhailuo: Added

PR #69, opened 2019-12-17
REG-142 add regulome summary download

yunhailuo:


yunhailuo:
> Looks good, though I do have a question about the .btn class on the parent `<div>`. Sort of makes it looks like two child buttons inside a parent button, but wasn’t sure if styling that parent div with .btn had a benefit.
> 
> I only looked at the Javascript part; sounds like the Python part is done.

Thank you Forrest. Fixed in a new commit.

PR #70, opened 2019-12-20
REG-171 add reset for browser

emro:
Adds a reset button to the genome browser.
Also added a `key` to prevent console error.

### DIFF
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -549,11 +549,10 @@ class FacetButton extends React.Component {
     }
 
     render() {
-        const { buttonLabel, lookupFilterCount, selectedFacets, facetLabel, key } = this.props;
+        const { buttonLabel, lookupFilterCount, selectedFacets, facetLabel } = this.props;
         const filterCount = lookupFilterCount(buttonLabel, facetLabel);
         return (
             <button
-                key={key}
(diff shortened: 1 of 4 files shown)


src/encoded/static/components/regulome_search.js
```diff
@@ -549,10 +549,11 @@ class FacetButton extends React.Component {
     }
 
     render() {
-        const { buttonLabel, lookupFilterCount, selectedFacets, facetLabel } = this.props;
+        const { buttonLabel, lookupFilterCount, selectedFacets, facetLabel, key } = this.props;
         const filterCount = lookupFilterCount(buttonLabel, facetLabel);
         return (
             <button
+                key={key}
```
forresttanaka: I don’t think you need to have the key here since this isn’t immediately in a loop. Just adding the key below in `Facet` like you have now should be fine.

src/encoded/static/components/regulome_search.js
```diff
@@ -587,6 +589,7 @@ const Facet = (props) => {
                         buttonLabel={d}
                         facetLabel={facetName}
                         addGenomeFilter={addGenomeFilter}
+                        key={facetName}
```
forresttanaka: …except you need to have something unique for each loop iteration here, so something unique based on each element of `facetArray` should fix the issue. `facetName` is the same for every iteration, so this I think would normally generate a console error.

PR #75, opened 2020-01-24
REG-173 accessibility page redesign

emro:
On the accessibility tab, we're replacing the lists of results with tables, and moving the tables to the left so there is less confusing white space.
We're keeping the lists of results for mobile, however, since the tables are too wide for good mobile viewing.

### DIFF
--- a/src/encoded/static/components/visualizations.js
+++ b/src/encoded/static/components/visualizations.js
@@ -605,9 +605,6 @@ export class ChartList extends React.Component {
                                             <p><span className="table-label">Organ</span>{d2.biosample_ontology.organ_slims.join(', ')}</p>
                                         : null}
                                         <p><span className="table-label">Method</span>{d2.method}</p>
-                                        {d2.biosample_ontology ?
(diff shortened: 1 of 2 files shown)


forresttanaka:
Starting to wonder, would it be worth looking into [matchMedia](https://www.sitepoint.com/javascript-media-queries/)? Browser support is good, and if the table is long (not sure how long it could get), that could avoid rendering the same data twice, once hidden and once not.

emro:
It's a good point that I should be avoiding rendering the same data twice. I could toggle the display based on `window.innerWidth` and then update on resize. Would using matchMedia be better than checking the window width?

forresttanaka:
Either way should work I think (I’ve never used matchMedia). It sounds like the advantage of matchMedia is that the Javascript would only be called when the window gets resized to a CSS breakpoint, where resize events would get called continually as you resize the window. So performance with matchMedia could in theory be better.

emro:
Maybe we can accomplish the same thing as matchMedia by using Lodash's debounce function.

emro:
@forresttanaka I think it is all set now!

PR #76, opened 2020-01-27
REG-161-add-marked-location

emro:
We've added an update to the Valis library which allows it to display a highlight location. This helps the user not lose track of the initial position of the browser. This branch points to the Valis update and adds the parameters to the browser configuration that will display the highlight.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "dalliance": "git+https://github.com/ENCODE-DCC/dalliance.git#v1.0",
     "domready": "^0.3.0",
     "form-serialize": "^0.6.0",
-    "genome-visualizer": "github:VALIS-software/valis-hpgv",
+    "genome-visualizer": "git+https://github.com/ENCODE-DCC/valis-hpgv.git#REG-161-add-marked-location",
     "google-analytics": "file:node_shims/google-analytics",
     "immutable": "^3.7.5",
     "jsonschema": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "dalliance": "git+https://github.com/ENCODE-DCC/dalliance.git#v1.0",
     "domready": "^0.3.0",
     "form-serialize": "^0.6.0",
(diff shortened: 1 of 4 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -359,9 +359,13 @@ class GenomeBrowser extends React.Component {
     }
 
     drawTracks(container) {
+        const highlightLocation = Math.floor((this.state.x1 - this.state.x0) / 2) + this.state.x0;
```
forresttanaka: Mathematically the same as avg(x1, x0) (not that JS has an average function, surprisingly). Would it be clearer that this averages if you have `Math.floor((this.state.x1 + this.state.x0) / 2)`, or does the addition of x0 outside Math.floor have an important effect?

emro: Oh yes, I could write it like this. I'm not entirely sure why I wrote it the other way.

package.json
```diff
@@ -72,7 +72,7 @@
     "dalliance": "git+https://github.com/ENCODE-DCC/dalliance.git#v1.0",
     "domready": "^0.3.0",
     "form-serialize": "^0.6.0",
-    "genome-visualizer": "github:VALIS-software/valis-hpgv",
+    "genome-visualizer": "git+https://github.com/ENCODE-DCC/valis-hpgv.git",
```
forresttanaka: I had a question about this. It looked like the old line pointed to the original Valis repo, but somehow we use the one you forked? Ben said we use “ours.”

hitz: we forked it and now we use that one (until this PR they were identical)

Surprised no one caught the stray dalliance dependency!  That should go.  (Eventually)

package.json
```diff
@@ -72,7 +72,7 @@
     "dalliance": "git+https://github.com/ENCODE-DCC/dalliance.git#v1.0",
```
hitz: remove!

package.json
emro: I've removed it! Thank you for catching that.

PR #77, opened 2020-01-29
REG-178 fix motif error

emro:
The sequence logo heights were not being computed quite correctly, the code was using a maximum instead of a summation to compute the relative frequencies.

### DIFF
--- a/src/encoded/static/libs/d3-sequence-logo.js
+++ b/src/encoded/static/libs/d3-sequence-logo.js
@@ -84,6 +84,7 @@ function offsets(counts, maxCount) {
     let H = 0;
     const sumCounts = counts.reduce((a, b) => +a + +b, 0);
     counts.forEach((d) => {
+        // we are adding an epsilon of 0.25 because that is consistent with the University of Michigan's implementation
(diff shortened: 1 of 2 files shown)


src/encoded/static/libs/d3-sequence-logo.js
```diff
@@ -76,30 +76,33 @@ function getLetterPath(i, strand) {
...
+    counts.forEach((d, j) => {
+        // relative frequency for base at position i
+        const relativeFrequency = (d + 0.25) / sumCounts;
+        // height of letter
+        const height = relativeFrequency * R;
+
+        // calculate offset from the top
+        // (maxCount / 2) is for the purposes of 'entryPoint'
+        // 'entryPoint' scales each letter by the max PWM value
+        // therefore we need to here we translate our values in the range [0,2] to be in the range [0, maxCount]
```
forresttanaka: “…we need to translate…”?

emro: Oops! I also took this opportunity to add in a comment about the 0.25 which is an addition that is specific to our implementation.

PR #78, opened 2020-01-30
REG-14 update dbSNP reference

yunhailuo:


src/encoded/regulome_indexer.py
```diff
@@ -258,9 +254,6 @@ def get_snp_index_mapping(chrom='chr1'):
                 'coordinates': {
                     'type': 'integer_range'
                 },
-                'strand': {
```
hitz: I guess SNPs don't have a strand?  Or maybe this code is left over from reading it from a BED

src/encoded/regulome_indexer.py
```diff
@@ -904,7 +891,7 @@ def metadata_doc(afile, dataset):
             if dataset_type in dataset['@type']:
                 meta_doc['dataset_type'] = dataset_type
                 break
-        if afile.get('submitted_file_name') in SNP_FILES:
+        if dataset_type == 'Reference':
```
hitz: This seems a little sloppy; there are lots of references that might contain a parsable/indexable file that aren't SNPs.  Not sure I would worry about it now though.

src/encoded/regulome_indexer.py
yunhailuo: Right. dbSNP data don't have strand info.

src/encoded/regulome_indexer.py
yunhailuo: When it's here, it should already been filtered above with internal_tag (line 828-835). I only tag SNP reference as RegulomeDB. It's still not a very good solution. Any suggestions?

PR #80, opened 2020-02-04
REG-180 fix pip version

emro:
We don't actually want this PR, it is a test.


emro:
I just opened this PR to test integration with Jira.

PR #82, opened 2020-02-19
REG-182 add link to legacy site

emro:
Adding a link to the legacy Regulome website.


src/encoded/static/scss/encoded/modules/_regulome_search.scss
```diff
@@ -1498,4 +1498,9 @@ button {
             padding-right: 3px;
         }
     }
+}
+
+.link-to-legacy-regulome {
+    text-align: center;
+    padding: 100px 20px 20px;
 }
```
forresttanaka: Carriage return at the end

PR #84, opened 2020-02-24
REG-183 remove beta banner

emro:
Removing the "beta" banner from the upper right corner.


emro:
I think this almost merits its own component, but it would be a very small component. If anything else gets added to the logo, I'll make this change.

PR #91, opened 2020-05-02
REG-196-parallelize-regulome-search-script

yunhailuo:


src/encoded/commands/regulome_search.py
```diff
@@ -14,73 +16,100 @@
...
-                result['score'] = atlas.regulome_score(
+            if not self.matched_pwm_peak_bed_only:
+                result['score'] = self.atlas.regulome_score(
                     all_hits['datasets'], evidence
                 )
                 result['features'] = evidence_to_features(evidence)
         except Exception:
-            logging.error(
-                'Regulome search failed on {}:{}-{}'.format(chrom, start, end)
+            return 1, 'Regulome search failed on {}:{}-{}'.format(
```
hitz, line 107: is 1 the correct return code here?

yunhailuo:
> I think I would probably use '0' for fail, and '1' for success... but maybe you have a good reason?? UNIX return codes?

No good reason. Just thought it's common? https://en.wikipedia.org/wiki/Exit_status

src/encoded/commands/regulome_search.py
yunhailuo: It failed here; so I return 1 as failed.

PR #93, opened 2020-06-16
REG-199-index-filter-MAF

yunhailuo:


src/encoded/regulome_indexer.py
```diff
@@ -1139,7 +1180,7 @@ def add_file_to_regions_es(self, request, afile, file_doc, snp=False):
                         continue  # Skip for 63 invalid peak in a non-ENCODE ChIP-seq result, exo_HelaS3.CTCF.bed.gz
                     if chrom not in SUPPORTED_CHROMOSOMES:
                         continue   # TEMPORARY: limit both SNPs and regions to major chroms
-                    if chrom not in file_data:
+                    if (chrom not in file_data) or (len(file_data[chrom]) > 3e6):
```
hitz, line 1183: This seems a little arbitrary, what is going on here?  3,000,000?

src/encoded/regulome_search.py
```diff
@@ -330,20 +332,33 @@ def parse_region_query(request):
...
             if (int(end) - int(start)) > 1:
                 notifications[region_query] = (
-                    'Failed: no known variants found in this multi-nucleotide '
-                    'region.'
+                    'Failed: no known variants matching query conditions found.'
                 )
                 continue
             else:
-                variants['{}:{}-{}'.format(chrom, int(start), int(end))] = set()
+                variants[(chrom, int(start), int(end))] = set()
```
hitz, line 359: Did you get rid of this so you could sort it or something?

src/encoded/regulome_search.py
```diff
@@ -374,7 +389,15 @@ def parse_region_query(request):
         'format': format,
         'from': from_,
         'total': total,
-        'variants': {k: list(v) for k, v in sorted(variants.items())[from_:to_]},
+        'variants': [
+            {
+                'chrom': chrom,
+                'start': start,
+                'end': end,
+                'rsids': sorted(variants[(chrom, start, end)])
```
hitz, line 397: ahh... here.

src/encoded/regulome_search.py
```diff
@@ -450,39 +478,35 @@ def regulome_summary(context, request):
...
-                'end': end,
-                'rsids': rsids,
-                'features': features,
-                'regulome_score': regulome_score
-            })
-        result['timing'].append({coord: (time.time() - begin)})  # DEBUG timing
+            variant['features'] = features
+            variant['regulome_score'] = regulome_score
+        result['timing'].append(
+            {'{}:{}-{}'.format(chrom, start, end): (time.time() - begin)}
```
hitz, line 499: Now putting back in the formatting...

src/encoded/regulome_search.py
```diff
@@ -413,8 +436,13 @@ def regulome_summary(context, request):
 
     # Redirect to regulome report for single unique region query
     if len(result['variants']) == 1:
-        query = {'regions': [v for v in result['variants']][0],
-                 'genome': result['assembly']}
+        query = {
+            'regions': [
+                '{}:{}-{}'.format(v['chrom'], v['start'], v['end'])
```
hitz, line 441: why does the query expect this format?  I guess that's what the UI sends it.

src/encoded/static/components/regulome_search.js
```diff
@@ -282,6 +283,7 @@ class AdvSearch extends React.Component {
 
                             <input type="submit" value="Search" className="btn btn-sm btn-info" />
                             <input type="hidden" name="genome" value={this.state.genome} />
+                            <input type="hidden" name="maf" value={this.state.maf} />
```
zoldello, line 286: I have not see hidden input type in a long time.

src/encoded/static/components/regulome_search.js
```diff
@@ -290,8 +292,8 @@ class AdvSearch extends React.Component {
                 {(context.notification && context.notification !== 'No annotations found') ?
                     <div className="notification">{context.notification}</div>
                 : null}
-                {(context.variants && context.variants.length > 0) ?
-                    <p>Searched coordinates: {Object.keys(context.variants)[0]}</p>
+                {(context.query_coordinates && context.query_coordinates.length > 0) ?
```
zoldello, line 295: Have you considered saving context.query_coordinates in a variable?

src/encoded/static/components/regulome_search.js
```diff
@@ -341,7 +343,7 @@ const NearbySNPsDrawing = (props) => {
     if (context.nearby_snps && context.nearby_snps[0] && context.nearby_snps[0].coordinates) {
         startNearbySnps = +context.nearby_snps[0].coordinates.lt;
         endNearbySnps = +context.nearby_snps[context.nearby_snps.length - 1].coordinates.lt;
-        coordinate = +Object.keys(context.variants)[0].split('-')[1];
+        coordinate = +context.query_coordinates[0].split('-')[1];
```
zoldello, line 346: I understand you using this to maintain consistency, but it is quite confusing. It is normally written as: 
coordinate =+ context.query_coordinates[0].split('-')[1];

src/encoded/static/components/regulome_search.js
```diff
@@ -1252,14 +1272,50 @@ class RegulomeSearch extends React.Component {
...
+        const coordinates = context.query_coordinates[0];
         const allData = context['@graph'] || [];
         const QTLData = allData.filter(d => (d.method && d.method.indexOf('QTL') !== -1));
         const chipData = allData.filter(d => d.method === 'ChIP-seq');
         const dnaseData = allData.filter(d => (d.method === 'FAIRE-seq' || d.method === 'DNase-seq'));
         const chromatinData = allData.filter(d => (d.method === 'chromatin state'));
         const thumbnail = this.context.location_href.split('/thumbnail=')[1] || null;
 
+        const toggleFreqsShow = () => this.setState(
+            state => ({ showMoreFreqs: !state.showMoreFreqs })
```
zoldello, line 1284: Let's talk about this.

src/encoded/static/components/regulome_search.js
```diff
@@ -1252,14 +1272,50 @@ class RegulomeSearch extends React.Component {
...
+        );
+        const hitSnps = {};
+        const sortedPopulations = {};
+        if (coordinates) {
+            const [chrom, startEnd] = coordinates.split(':');
+            const [start, end] = startEnd.split('-');
+            context.nearby_snps.forEach((snp) => {
+                if (snp.chrom === chrom && snp.coordinates.gte === +start && snp.coordinates.lt === +end) {
```
zoldello, line 1292: Let's talk about this.

src/encoded/static/components/regulome_search.js
```diff
@@ -1252,14 +1272,50 @@ class RegulomeSearch extends React.Component {
...
+                            });
+                        });
+                    }
+                    if (snp.alt_allele_freq) {
+                        Object.keys(snp.alt_allele_freq).forEach((allele) => {
+                            Object.keys(snp.alt_allele_freq[allele]).forEach((population) => {
+                                if (!hitSnps[snp.rsid][population]) {
+                                    hitSnps[snp.rsid][population] = `${allele}=${snp.alt_allele_freq[allele][population]}`;
+                                } else {
+                                    hitSnps[snp.rsid][population] += `, ${allele}=${snp.alt_allele_freq[allele][population]}`;
```
zoldello, line 1307: why not use join? Here is an example:
['a', 'b'].join(' , ')
As in python, in Javascript, strings are immutatable. So, you are constantly creating a new string and its a slow process.

src/encoded/static/components/regulome_search.js
```diff
@@ -1305,6 +1361,26 @@ class RegulomeSearch extends React.Component {
...
+                                {Object.keys(hitSnps).map(rsid =>
+                                    <div className="notification-line" key={rsid}>
+                                        <div className="notification-label">{rsid}</div>
+                                        <div className="notification">
+                                            <div>
+                                                {sortedPopulations[rsid].slice(0, 3).map(
+                                                    population => <div>{`${hitSnps[rsid][population]} (${population})`}</div>
+                                                )}
+                                            </div>
+                                            {sortedPopulations[rsid].length > 3 && this.state.showMoreFreqs ?
```
zoldello, line 1373: Why 3?

src/encoded/regulome_indexer.py
yunhailuo: It is arbitrary. If we hold all SNP docs of one chromosome in memory before putting into index, it will go out of memory. Any suggestions on better strategies here?

src/encoded/regulome_search.py
yunhailuo: Yes, here. Honestly, this doesn't have to be sorted. I just don't like arbitrary order. Maybe we should go back?

src/encoded/regulome_search.py
yunhailuo: It can be anything which can be parsed properly here: https://github.com/ENCODE-DCC/regulome-encoded/blob/dev/src/encoded/regulome_search.py#L258-L279 But I assume it has to be a string since it's redirected to the other endpoint. Maybe I'm wrong?

src/encoded/static/components/regulome_search.js
yunhailuo: I just copied whatever Emma did here: https://github.com/ENCODE-DCC/regulome-encoded/blob/dev/src/encoded/static/components/regulome_search.js#L284

src/encoded/static/components/regulome_search.js
yunhailuo: No. Maybe there is an advantage? Again I was trying to mimic the behavior of `context.variants`.

src/encoded/static/components/regulome_search.js
yunhailuo: Should I fix others up there to really maintain consistency?

src/encoded/static/components/regulome_search.js
yunhailuo: Thanks for explaining that. Make sense. I can build an array and join after the iteration.

src/encoded/static/components/regulome_search.js
yunhailuo: In general, arbitrary. Specifically, I got this "feeling" from dbSNP like this: https://www.ncbi.nlm.nih.gov/snp/rs3768324. Any good suggestions?

src/encoded/regulome_indexer.py
hitz: at least define a constant like "MAX_SNP_FILE = 3e6",  better would be to stream the file using IOStream.  Not sure this is worth editing.

src/encoded/regulome_search.py
hitz: I guess it's OK.

src/encoded/static/components/regulome_search.js
hitz: 3 is fine, but naked integer parameter is a little wonky.  Not sure what the protocol is for JS constants.

PR #104, opened 2020-10-10
REG-209-script-search-snps-in-region

yunhailuo:


hitz (commented):
My only concern is that I don't know if there is anything in place to clean up Temporary Files.  Just because they are in /tmp (or whatever) doesn't mean that there something to clean up.  Probably they are tiny and don't matter, but have to be a little careful if this is running on a production server.

yunhailuo:
> My only concern is that I don't know if there is anything in place to clean up Temporary Files. Just because they are in /tmp (or whatever) doesn't mean that there something to clean up. Probably they are tiny and don't matter, but have to be a little careful if this is running on a production server.

I assume it will get clean up: https://docs.python.org/3/library/tempfile.html

> This module creates temporary files and directories. It works on all supported platforms. TemporaryFile, NamedTemporaryFile, TemporaryDirectory, and SpooledTemporaryFile are high-level interfaces which provide automatic cleanup and can be used as context managers. mkstemp() and mkdtemp() are lower-level functions which require manual cleanup.

hitz (approved):

PR #112, opened 2021-06-25
REG-198-add-download-button

emro:
- Adds a download button for ENCFF297XMQ.tsv file on Encode
- Adds new options/styles for batch download modals
- Minor fix: "less" should be "fewer"


pedrohr (approved):
👍

PR #116, opened 2021-07-30
REG-228 fix browser crash

emro:
- Updated variable names (target.label) to prevent browser crashing
- Rewrote promises code


pedrohr (approved):
💯

pedrohr (approved):

PR #124, opened 2021-10-22
REG-237-clean-cloud-config

ottojolanki:
Remove package installations that are now already done in the AMI. We do not need to configure postfix, it is not installed anymore.

### DIFF
--- a/cloud-config.yml
+++ b/cloud-config.yml
@@ -3,47 +3,7 @@ ssh_authorized_keys:
   - %(LOCAL_SSH_KEY)s
 bootcmd:
 - set -ex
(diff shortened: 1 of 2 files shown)


stuartmiyasato (approved):
I think this is fine, although i'm not exactly clear about the purpose of lines 14-16?  It seems like line 16 basically removes what is done in lines 14-15, so it looks unnecessary to me.  But it also doesn't seem to hurt anything, so I think it's okay.

stuartmiyasato (approved):
Looks good to me.

stuartmiyasato (approved):
Looks good to me.

PR #128, opened 2022-02-01
rectangular highlight location

leepc12:
changed highlight location string format from `contig:start` to `contig:start-end`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "dagre-d3": "git://github.com/ENCODE-DCC/dagre-d3.git#use-local-d3",
     "domready": "^0.3.0",
     "form-serialize": "^0.6.0",
-    "genome-visualizer": "git://github.com/ENCODE-DCC/valis-hpgv.git#1.8.0",
+    "genome-visualizer": "ENCODE-DCC/valis-hpgv#bca601848038bfdc39bcc8090e37da3c1067e5ab",
     "google-analytics": "file:node_shims/google-analytics",
     "immutable": "^3.7.5",
     "jsonschema": "^1.1.0",
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -319,7 +319,10 @@ TrackLabel.propTypes = {
 class GenomeBrowser extends React.Component {
     constructor(props, context) {
         super(props, context);
-
+        const highlightLocationStart = +props.coordinates.split(':')[1].split('-')[0];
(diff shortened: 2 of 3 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -319,16 +319,21 @@ TrackLabel.propTypes = {
 class GenomeBrowser extends React.Component {
     constructor(props, context) {
         super(props, context);
-
+        const highlightLocationStart = +props.coordinates.split(':')[1].split('-')[0];
+        const highlightLocationEnd = +props.coordinates.split(':')[1].split('-')[1];
+        const x0 = highlightLocationStart - 2000;
+        const x1 = highlightLocationEnd + 2000;
```
emro: This should probably be 5000, or if we leave it, Ben should confirm that's ok.

src/encoded/static/components/genome_browser.js
```diff
@@ -563,8 +568,9 @@ class GenomeBrowser extends React.Component {
     }
 
     drawTracks(container) {
-        const highlightLocation = Math.floor((this.state.x1 + this.state.x0) / 2) + 1; // browser uses base 1
-        const highlightString = `${this.state.contig}:${highlightLocation}`;
+        const highlightLocationStart = this.state.highlightLocationStart;
+        const highlightLocationEnd = this.state.highlightLocationEnd;
+        const highlightString = `${this.state.contig}:${highlightLocationStart}-${highlightLocationEnd}`;
```
emro, line 573: I would probably have skipped lines 571 and 572 and just used this.state in line 573 but this doesn't matter. I think our linting program on Encode would probably suggest destructuring here.

emro (commented):
Ok if Ben says the change in displayed range is ok.

src/encoded/static/components/genome_browser.js
emro: @hitz says this is ok

emro (approved):

package.json
```diff
@@ -71,7 +71,7 @@
     "dagre-d3": "git://github.com/ENCODE-DCC/dagre-d3.git#use-local-d3",
     "domready": "^0.3.0",
     "form-serialize": "^0.6.0",
-    "genome-visualizer": "git://github.com/ENCODE-DCC/valis-hpgv.git#1.8.0",
+    "genome-visualizer": "ENCODE-DCC/valis-hpgv#bca601848038bfdc39bcc8090e37da3c1067e5ab",
```
hitz, line 74: this line should be changed to valis-hpgv.git#1.9.0

PR #136, opened 2022-02-23
REGD-36-docker

mingjiecn:


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN apt-get -y install openssl
+RUN apt-get -y install graphviz
+RUN apt-get -y install nginx
+RUN apt-get -y install freetype2-demos
+RUN apt-get -y install libjpeg-dev
+RUN apt-get -y install libtiff-dev
+RUN apt-get -y install liblcms2-utils
+RUN apt-get -y install webp
+RUN apt-get -y install nodejs
+RUN apt-get -y install npm
```
ottojolanki: It would be better to put all these in one RUN command to avoid creating multiple layers. We can just do 
```
apt -get y- install pkg1 \
        pkg2 \
        pkg3 
```
etc

README.md
```diff
@@ -137,6 +137,22 @@ pyenv local 3.8.10
...
+  - `docker build . -t regulome`
+
+### **2. running the app using docker or docker compose**  
+
+- You can use docker to run regulome:
+  - `docker run -d -p 6543:6543 regulome`
+
+- You can also use docker compose to run regulome:
+  - `docker-compose up`
+
```
ottojolanki, line 154: This runs into error:
```
$ docker-compose up --build
[+] Running 0/1
 ⠿ regulome Error                                                                                                                                                                                                                                1.5s
Error response from daemon: pull access denied for regulome, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

ottojolanki: See comment on `docker-compose.yml`

docker-compose.yml
```diff
@@ -0,0 +1,7 @@
+version: '3'
+
+services:
+  regulome:
+    image: regulome
```
ottojolanki: I think `image` key is meant to be used for images that are available in some online registry. For local stuff you can replace this with `build: .`.

ottojolanki (changes requested):
Compose file and Dockerfile require small tweaks.

ottojolanki (approved):
LGTM

PR #137, opened 2022-02-23
REGD-4 new logo format

emro:
- Adds in file reader for new motif data format (see Jira ticket for details about format)
- Fixes duplicate motifs display bug
- Fixes error when browser attempts to draw motifs before component is mounted


mingjiecn (approved):
I wasn't able to see the error in the console, but the duplication is fixed.

PR #138, opened 2022-03-14
REGD-38 hide download buttons for no results

emro:
When no SNP is specified or an invalid SNP(s) (and no valid SNPs) is specified, a pop-up with an error message is displayed instead of the page navigating to the summary page with no results.


mingjiecn (approved):
It is working.

PR #139, opened 2022-03-16
REGD-39-remove-dependency-dagre-d3-change-git-old-protocol

mingjiecn:


emro (approved):
Looks good!

We often follow the convention of the first commit having the ticket number but that's fine.

PR #141, opened 2022-03-28
REGD-41-docker-dev-container

mingjiecn:


ottojolanki:
Maybe now that we can run the tests using docker, we should move the circleCI tests to docker as well?

ottojolanki (approved):
Nice work! LGTM.

PR #145, opened 2022-05-27
REGD-43 improve chromatin states page

emro:
- Adds body map to chromatin view (also updates the body map to not rely on url as on Encode)
- Colors the body map based on most active chromatin state with result for each organ
- Removes bars for chromatin state filter
- Adds biosample filter ordered by organ
- Updates table of results to be sortable and display classification column
- Reorganizes code so facet-related components/functions are in a separate file
- Updates colors for chromatin states


src/encoded/static/img/bodyMap/insetSVGs/lymphatic_vessel.svg
```diff
@@ -0,0 +1,7 @@
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 191 191" style="enable-background:new 0 0 191 191;" xml:space="preserve">
+<image style="overflow:visible;enable-background:new    ;" width="191" height="191" xlink:href="96A70511DFE00CCF.png" >
+</image>
+<image style="overflow:visible;" width="191" height="191" id="Layer_1_copy_00000045618743682790781740000010838555337368648379_" xlink:href="96A70511DFE00CCD.png" >
```
forresttanaka: I might not know how this works since I’ve never worked with an SVG that links to a PNG, but how does this link to (what I presume is) lymphatic_vessel.png? This 96A70511DFE00CCF.png file doesn’t seem to exist.

src/encoded/static/components/body_map.js
```diff
@@ -0,0 +1,758 @@
...
+// Pop-up body map facet
+// Displayed when you click on <ClickableThumbnail>
+// Allows you to select organ / system filters
+export const BodyMapModal = (props) => {
+    const { facet, isThumbnailExpanded, toggleThumbnail, organism, handleFilters, originalFilters } = props;
+    const [mapFacet, setFacet] = useState(facet);
+
+    useEffect(() => {
+        setFacet(props.facet);
+    });
```
forresttanaka: Does this not cause an setFacet to get called on every render even if props.facet doesn’t change? `useEffect` with no dependency array is defined to be called on every render.

src/encoded/static/components/body_map.js
```diff
@@ -0,0 +1,758 @@
...
+
+    const toggleThumbnail = () => {
+        setIsThumbnailExpanded(!isThumbnailExpanded);
+    };
+
+    React.useEffect(() => {
+        setFacet(props.facet);
+        // Highlight body map selections
+        initializeBodyMap(props.originalFilters, BodyList, props.facet.associatedStates);
+    });
```
forresttanaka: Another `useEffect` with no dependency array, which should cause it to get called even if not needed.

src/encoded/static/components/chromatin_view.js
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    const files = [];
+    filteredData.forEach((d) => {
+        files.push({
+            organ_slims: d.biosample_ontology.organ_slims,
+            biosample: d.biosample_ontology.term_name,
+            state: lookupChromatinNames(d.value),
+        });
+    });
```
forresttanaka: A .map() seems the more natural mechanism here.

src/encoded/static/components/chromatin_view.js
```diff
@@ -0,0 +1,399 @@
...
+                    label = keyArray.find(k => sanitizedString(k) === f);
+                } else if (filterType === 'state') {
+                    label = keyArray.find(k => classString(sanitizedString(k)) === f);
+                }
+                let clearInput = f;
+                if (filterType === 'biosample') {
+                    clearInput = sanitizedString(f);
+                } else if (filterType === 'chromatin') {
+                    clearInput = classString(sanitizedString(f));
+                }
```
forresttanaka, line 92: Can `filterType` ever contain a value other than the ones you check here?

src/encoded/static/components/chromatin_view.js
```diff
@@ -0,0 +1,399 @@
...
+        super(props);
+
+        const files = [];
+        this.props.data.forEach((d) => {
+            files.push({
+                organ_slims: d.biosample_ontology.organ_slims,
+                biosample: d.biosample_ontology.term_name,
+                state: lookupChromatinNames(d.value),
+            });
+        });
```
forresttanaka: Similar thing as above. I would think .map() would be the more natural construct.

src/encoded/static/components/facets.js
```diff
@@ -0,0 +1,518 @@
...
+};
+
+const placeZerosAtEnd = (arr) => {
+    Object.filter = (obj, predicate) =>
+        Object.keys(obj)
+            .filter(key => predicate(obj[key]))
+            .reduce((res, key) => Object.assign(res, { [key]: obj[key] }), {});
+
+    const isntZero = Object.filter(arr, val => val > 0);
+    const isZero = Object.filter(arr, val => val === 0);
```
forresttanaka: I guess `arr` isn’t an array? `Object.filter` gets redefined every time `placeZerosAtEnd` gets called, so it seems like you could define this function outside of `placeZerosAtEnd`. Highly unusual to add a function directly to the Object object. The usual way is to use prototypes, but I don’t see the benefit of even using prototypes here. I’m also not sure what `Object.filter` does, nor what the purpose of `placeZerosAtEnd` is.

src/encoded/static/components/facets.js
```diff
@@ -0,0 +1,518 @@
...
+                            addGenomeFilter={addGenomeFilter}
+                            key={d}
+                        />;
+                    })}
+                </div>
+                <div className={`shading ${(facetArray.length < displayedTermsCount) ? 'hide-shading' : ''}`} />
+            </div>
+        </div>
+    );
+};
```
forresttanaka, line 231: Seems like `TypeaheadFacet` could either call `Facet` directly to reduce duplication, or rename it and have it take a `children` property, and use `Facet` as a child of this new component, again to reduce the code duplication.

src/encoded/static/components/facets.js
```diff
@@ -0,0 +1,518 @@
...
+            const unsanitizedSearchTerms = { ...prevState.unsanitizedSearchTerms };
+            const searchTerms = { ...prevState.searchTerms };
+            unsanitizedSearchTerms[typeaheadIdentifier] = targetValue;
+            searchTerms[typeaheadIdentifier] = filterVal;
+            return {
+                searchTerms,
+                unsanitizedSearchTerms,
+            };
+        });
+    }
```
forresttanaka, line 340: I don’t think you need both `unsanitizedSearchTerms` _and_ `searchTerms` to be in state. Just calculate `searchTerms` when you need it, in `render()` — basically the philosophy Dan Abramov talks about here, in which there should be a high bar before putting anything in state: https://twitter.com/dan_abramov/status/749710501916139520?s=20&t=7-uNRnEgre6Im48YM-GtyQ

src/encoded/static/components/regulome_search.js
```diff
@@ -5,16 +5,20 @@ import url from 'url';
...
 
 const screenMediumMax = 787;
 const screenSmallMax = 483;
 
-// Number of terms to show, the rest will be viewable on scroll
-const displayedTermsCount = 14;
+// Define facets (probably this should be in a schema really)
+const facetList = ['file_format', 'organ', 'biosample', 'assay', 'target'];
+const facetTitleList = ['File format', 'Organ / cell type', 'Biosample', 'Method', 'Target'];
+const typeaheadFacetList = [false, true, true, false, true];
```
forresttanaka: Would it be more straightfoward to implement these as one array of objects, to make it easier to keep these in sync? I realize this is from existing code.

src/encoded/static/components/visualizations.js
```diff
@@ -793,205 +796,175 @@ export class ChartTable extends React.Component {
...
-                return obj;
-            }, {});
-        const newSelectedStates = [];
-        filteredData.forEach((d) => {
-            fakeFacets[lookupChromatinNames(d.value)] += 1;
-            const chromatinValue = sanitizedString(lookupChromatinNames(d.value));
-            if (!newSelectedStates.includes(chromatinValue)) {
-                newSelectedStates.push(chromatinValue);
+        if (prevProps.chartWidth !== this.props.chartWidth) {
+            if (this.props.chartWidth <= this.state.leftMargin) {
```
forresttanaka: Looks like `leftMargin` and  possibly `chartKeys` don’t need to be in state, but calculated in `render()`. AFAICT `chartMax` and `chartData` don’t change values unless props changes, so they don’t need to be in state either.

forresttanaka (changes requested):

src/encoded/static/img/bodyMap/insetSVGs/lymphatic_vessel.svg
emro: I'm actually not sure (I didn't make this file), but somehow it does work on Encode! That said, we're not actually using this svg or the lymph_node.svg file so they can be deleted.

src/encoded/static/components/chromatin_view.js
emro: Yes, it can be "biosample", "state" or "bodymap"

src/encoded/static/components/visualizations.js
emro: Ok I re-wrote all this. It was pretty old code and needed some revisions.

src/encoded/static/components/facets.js
emro: Yes, this is all pretty old and slightly weird code that I slightly rewrote. It's not actually used on the chromatin view page, but I thought I'd move the facet related code out of regulome_search.js and put it with the other code used to generate facets. This is only used on the genome browser view.

I added some comments, but it basically displays all of the terms with 0 results at the bottom of the browser facets. Unlike in Encode, we want to show these, we just want them to be disabled.

Screen Shot 2022-06-01 at 11 40 17 AM

This is where I found this strategy for filtering objects: https://stackoverflow.com/questions/5072136/javascript-filter-for-objects

src/encoded/static/components/visualizations.js
emro: Sometimes I think that most of Regulome could stand to be re-written, maybe I'll get to it piece by piece as we redo the pages.

emro:
I realized there was another really important change that I needed to make. I'd re-written the genome browser facet data structure and code such that you could not make multiple selections in situations where you should be able to. For instance, you should be able to select both bigWig and bigBed files, and not have bigBeds disabled after a selection of bigWigs. I reverted some of my changes and the genome browser facets now work the way they do on production.

forresttanaka (approved):
Yeah, looks pretty clean.

src/encoded/static/components/facets.js
```diff
@@ -7,6 +7,14 @@ import { initializedChromatinObject } from './chromatin_view';
 // Number of terms to show, the rest will be viewable on scroll
 const displayedTermsCount = 14;
 
+// Add method to filter objects based on value
+Object.filter = (obj, predicate) =>
+    Object.keys(obj)
+        .filter(key => predicate(obj[key]))
+        .reduce((res, key) => Object.assign(res, { [key]: obj[key] }), {});
```
forresttanaka, line 14: Interesting. I didn’t know extending Object with prototype was frowned upon, though the page they linked to didn’t really say that, or maybe I misunderstood (not that I would extend the prototype of Object in my own code). Anyway, I don’t see the advantage of this over just a regular function to do the filtering, as the poster also said, “Instead, provide your feature as stand-alone function, or if you really want to extend a global, provide it as utility function on Object.”

forresttanaka (approved):

PR #148, opened 2022-07-18
REGD-51 update to GRCh38

emro:
Makes changes to update the site to GRCh38:
- default assembly is now GRCh38 and GRCh37 is renamed hg19 everywhere as that is how the files are labeled
- switch is added on /regulome-search/ to allow toggle between assemblies
- browser page
-- pinned files are updated
-- preferred default is used to get display files
-- we now show the file and dataset names when the tracks expand
-- bigBeds can now expand
-- browser links use AWS bucket instead of encodeproject.org link and that should fix Safari track display bug
- ChIP page
-- long charts should be scrollable and readable now (short charts look a little silly but this has to be redesigned anyway)
- chromatin page
-- use GRCh38 state model (update names and colors), re-order how the states are displayed
-- use cutoff to determine which facet names need extra height instead of hard-coded list
- accessibility page
-- rename titles to include only DNase-seq
- motifs and footprints page
-- no changes required on front end
- QTL data
-- limit very long charts on thumbnail to 10 rows


src/encoded/static/components/chromatin_view.js
```diff
@@ -1,37 +1,58 @@
...
 
 const sanitizedString = globals.sanitizedString;
 const classString = globals.classString;
 
-export const initializedChromatinObject = {
+export const initializedChromatinObjectHg19 = {
     'Active TSS': 0,
     'Flanking Active TSS': 0,
+    'Genic enhancers': 0,
+    Enhancers: 0,
```
forresttanaka, line 16: As a side note, in igvf-ui I didn’t enable the rule requiring no quotes around single-word object keys, so you can put quotes on everything if you feel like it — space-containing or not. That just seems visually nicer when looking at an object definition with both space-containing and non-space-containing keys.

src/encoded/static/components/chromatin_view.js
```diff
@@ -1,37 +1,58 @@
...
+    'Strong transcription': 0,
+    'Weak transcription': 0,
+    'Bivalent/Poised TSS': 0,
+    'Bivalent enhancer': 0,
+    'ZNF genes & repeats': 0,
+    'Weak Repressed PolyComb': 0,
+    'Repressed PolyComb': 0,
+    Heterochromatin: 0,
     'Quiescent/Low': 0,
 };
```
forresttanaka, line 49: It took me a while to understand why these weren’t arrays of strings since you nearly always get the keys anyway — _nearly_. I see you use it to initialize the fake facets. _Might_ be more straightforward to make these arrays of strings and then just use that to build the fake facets with 0 counts for all of them, but I don’t know which is more straightforward overall.

src/encoded/static/components/snps_diagram.js
```diff
@@ -0,0 +1,162 @@
...
+                        refY="5"
+                        markerWidth="5"
+                        markerHeight="6"
+                        orient="auto-start-reverse"
+                    >
+                        <path d="M 0 0 L 10 5 L 0 10 z" />
+                    </marker>
+                </defs>
+                <g className="grid x-grid" id="xGrid">
+                    <line x1="10" x2="990" y1="75" y2="75" markerEnd="url(#arrow)" markerStart="url(#arrow)" stroke="#7F7F7F" strokeWidth="2" />
```
forresttanaka, line 85: Would it be easier to maintain a global object of colors, and just use the key names for the strokes?

src/encoded/static/components/snps_diagram.js
```diff
@@ -0,0 +1,162 @@
...
+                            offset = 20;
+                        }
+                    } else {
+                        offset = 20;
+                    }
+                }
+            }
+            coordinateYOffset.push(offset);
+        });
+    }
```
forresttanaka, line 63: Phil liked to note magic numbers in the code, and while I usually let one or two obvious ones pass, this seems like a lot of fairly meaningless numbers.

src/encoded/static/components/visualizations.js
```diff
@@ -833,8 +865,18 @@ export const ChartTable = (props) => {
                         fullResultOrgan = (fullResult && fullResult.biosample_ontology) ? fullResult.biosample_ontology.organ_slims.join(', ') : '';
                     }
 
+                    let chromatinHierarchy = [];
+                    if (props.assembly && props.assembly === 'hg19') {
```
forresttanaka, line 869: Can use optional chaining here just to save on typing.

src/encoded/static/scss/encoded/modules/_chromatin_state.scss
```diff
@@ -380,19 +459,69 @@ $statestroke: #7D6167;
     }
 }
 
+.activeenhancer .overlay {
```
forresttanaka: The following area would have been perfect for a Sass @each loop — easier to maintain.

src/encoded/regulome_search.py
```diff
@@ -63,7 +64,7 @@ def regulome_search(context, request):
         return {
             '@context': '/terms/',
             '@id': '/regulome-search',
-            'assembly': 'GRCh37',
+            'assembly': 'GRCh38',
```
forresttanaka, line 67: Seems like everywhere else GRCh37 turned into hg19, but not here?

src/encoded/static/components/regulome_search.js
```diff
@@ -257,21 +258,42 @@ ExampleEntry.propTypes = {
...
+    <div className="assembly-switch">
+        <div className={`switch-label ${props.selection === 'GRCh38' ? 'active' : ''}`}>GRCh38</div>
+        <label className="switch" id="assembly-switch">
+            <input
+                type="checkbox"
+                onChange={() => { props.onSelection(props.selection); }}
+                checked={props.selection === 'hg19'}
+            />
+            <span className="slider round" />
+        </label>
```
forresttanaka: While I haven’t run this code, I’m guessing this checkbox isn’t accessible. The `<label>` doesn’t have speakable text in it.

forresttanaka (approved):
All my feedback are suggestions. It all looks like it works and is logical.

src/encoded/static/components/chromatin_view.js
emro: Yeah, this has always bothered me

src/encoded/static/components/snps_diagram.js
emro: I think we're going to get rid of this probably anyway, or do a lot of revisions (we want to show genes as well here) so I might postpone changes to it for now.

src/encoded/static/components/snps_diagram.js
emro: I think this will all get re-written. I nearly just deleted it all but I think we want to keep it until we have a replacement diagram.

src/encoded/regulome_search.py
emro: I'm not 100% confident about my python changes, but I think this is one of the two places where I'm trying to change the default to be GRCh38 instead of GRCh37/hg19. We want everything to default to GRCh38 now.

src/encoded/static/components/regulome_search.js
emro: Yes, this was not accessible and it was less trivial than I thought to make it tab-able since the input/checkbox is hidden. But I think it should be accessible now. Thank you for catching this. After I saw your comment I actually ran Lighthouse on this page and this wasn't flagged somehow. It definitely needed to be fixed.

src/encoded/static/components/chromatin_view.js
emro: Maybe I'll keep it as is for now

src/encoded/static/components/visualizations.js
emro: I always forget about this syntax

src/encoded/static/scss/encoded/modules/_chromatin_state.scss
emro: Yes, that would have saved some typing too

src/encoded/static/components/regulome_search.js
```diff
@@ -258,20 +258,36 @@ ExampleEntry.propTypes = {
...
+    // need to display a focus style on "switch-label" because the input default checkbox is hidden and focus is therefore hidden
+    const [focus, toggleFocus] = React.useState(false);
+
+    return (
+        <div className="assembly-switch">
+            <div className={`switch-label ${props.selection === 'GRCh38' ? 'active' : ''}`}>GRCh38</div>
+            <label
+                className={`switch ${focus ? 'focused' : ''}`}
+                htmlFor="assembly-switch"
+                aria-labelledby={`Assembly is ${props.selection}. Toggle to switch to ${props.selection === 'GRCh38' ? 'hg19' : 'GRCh38'}`}
```
forresttanaka: I think this should use `aria-label`. [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) only holds the id of another element.

forresttanaka (changes requested):
OK, all your comment replies sound good. Just a change to the aria-labelledby is needed.

emro:
I fixed the aria-label and updated the styles to make it more maintainable/concise. I tried to make the optional chaining update but our node version for Regulome doesn't support it.

forresttanaka (approved):
Yup, looks fine. I see what you mean about the three copies of colors. My understanding, unconfirmed by me, is that you can import Tailwind colors into Javascript, so that might help situations like that if true.

PR #150, opened 2022-08-09
more updates to REGD-50

dougli1sqrd:
* Target, Strand, and Pwm components separated for easier reading
* Target, Strand, and Pwm components also aligned into 3 columns on one line in each motif result
* The name of the Assembly Reference is set based on how the search is done (hg19 or GRCh38)
* A Black border surrounds the aligned sequence element in the reference to match the red rectangle in the sequence logo

### DIFF
--- a/src/encoded/static/components/motifs.js
+++ b/src/encoded/static/components/motifs.js
@@ -158,12 +158,27 @@ export class MotifElement extends React.Component {
         const targetList = element.targets;
         const footprintList = {};
         const pwmList = {};
+
+        /*
+        Element {
+            accessions [ ID ]
+            biosamples [ Label ]
+            datasets [ URL ]
+        }
+        */
         element.datasets.forEach((d, dIndex) => {
+            // (URL, Index)
+
             const biosample = element.biosamples[dIndex];
+            // biosample: Label
             const accession = element.accessions[dIndex];
+            // accession: ID
+            // So for a dataset d, we get the Label and ID corresponding to the same index we are at now
             if (biosample !== undefined) {
+                // If we have a biosample, then we map Label -> URL into `footprintList`
                 footprintList[biosample] = d;
             } else {
+                // If we don't have a biosample, then map Label -> URL into `pwmList`
                 pwmList[accession] = d;
             }
         });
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/motifs.js
```diff
@@ -175,16 +190,63 @@ export class MotifElement extends React.Component {
...
         const footprintsLabel = (footprintsLength > 1 || footprintsLength === 0) ? 'Footprints' : 'Footprint';
+        
+        const targetComponent = (
+            <React.Fragment>
+            {targetList.length > 0 && (
+                <React.Fragment>
+                    {targetListLabel} {targetList}
+                </React.Fragment>
+            )}
+            </React.Fragment>
```
forresttanaka, line 201: Normally `<React.Fragment>` gets treated as any other HTML element, so child elements would get indented like you have below in `strandComponent`.

src/encoded/static/components/motifs.js
```diff
@@ -158,12 +158,27 @@ export class MotifElement extends React.Component {
...
+
+        /*
+        Element {
+            accessions [ ID ]
+            biosamples [ Label ]
+            datasets [ URL ]
+        }
+        */
         element.datasets.forEach((d, dIndex) => {
+            // (URL, Index)
```
forresttanaka, line 170: The comment makes this much clearer. It would have been nice if they had named `d` and `dIndex` more clearly in the beginning. I initially configured igvf-ui to disallow single-character variable names, but it caused too many issues with npm packages.

forresttanaka (approved):
Only one thing to possibly change, but it’s not that big of a deal. Might need to change the “base” (is that what it’s called?) since it says you want to merge your branch into Emma’s REGD-50 branch. Not really sure how you do that. I’m also not really sure if there are more changes than I see here, hidden in Emma’s branch.

dougli1sqrd:
oh yeah, This is I guess for just my changes, and I assume there are other changes from Emma. I can try and push directly to the branch?

forresttanaka:
I’m not sure. Was the problem that you modified Emma’s branch but you couldn’t push it to GitHub?

forresttanaka:
This involved a _lot_ of changes, and many of them I couldn’t really tell what they were for, but I didn’t see anything specifically wrong, so I think just try it. The changes @dougli1sqrd looked fine to me.

PR #151, opened 2022-08-10
REGD-50 motif panel updates

dougli1sqrd:
Motif panel updates

* Motif search results take up a little less vertical space by making one row of 3 columns for info that was all vertical
* Reference sequence has a highlight on the centered element
* Search result title changes name properly based on the assembly reference: hg19 or GRCh38
* And other changes


src/encoded/static/components/chromatin_view.js
```diff
@@ -82,7 +103,7 @@ const Selections = (props) => {
                 if (filterType === 'biosample') {
                     label = keyArray.find(k => sanitizedString(k) === f);
                 } else if (filterType === 'state') {
-                    label = keyArray.find(k => classString(sanitizedString(k)) === f);
+                    label = keyArray.find(k => (!isLetter(k[0]) ? (classString(sanitizedString(k)) === f) : (sanitizedString(k) === f)));
```
forresttanaka: Not really sure what this is for, especially with the data involved not commented. I’ll have to assume it’s proper.

src/encoded/static/components/motifs.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import _ from 'underscore';
 import PropTypes from 'prop-types';
-// import * as logos from '../libs/d3-sequence-logo'; // This is for local development when changes are needed to d3-sequence-logo.
+import * as logos from '../libs/d3-sequence-logo'; // This is for local development when changes are needed to d3-sequence-logo.
```
forresttanaka: I guess this means this needs to be commented out before merging to dev?

src/encoded/static/components/motifs.js
```diff
@@ -287,6 +293,49 @@ export const Motifs = (props) => {
...
+                } else if (nucleotide === 'G') {
+                    fakePWM.push([0, 0, maxHeight, 0]);
+                } else {
+                    fakePWM.push([0, 0, 0, maxHeight]);
+                }
+            });
+
+            sequenceLogos.entryPoint(refToReference.current, fakePWM, d3lib, referenceStart, referenceStart, referenceStart, '+', true);
+        });
+    });
```
forresttanaka: This has no dependency array, so the callback gets called every render. Is that supposed to happen (sometimes that’s valid, but that seems unusual). Also the scroll handler doesn’t get removed when this component unmounts.

src/encoded/static/components/motifs.js
```diff
@@ -175,38 +190,80 @@ export class MotifElement extends React.Component {
...
         const footprintsLabel = (footprintsLength > 1 || footprintsLength === 0) ? 'Footprints' : 'Footprint';
+        
+        const targetComponent = (
+            <React.Fragment>
+            {targetList.length > 0 && (
+                <React.Fragment>
+                    {targetListLabel} {targetList}
+                </React.Fragment>
+            )}
+            </React.Fragment>
```
forresttanaka, line 200: Non-standard indenting.

forresttanaka (changes requested):
Mostly I think the event listener in the useEffect should get removed when the component gets unmounted, but also I’d like to make sure that not having the dependency array on that useEffect is intended.

There’s also a lot of commented-out code which usually gets dinged in code reviews. I think Regulome might have a sort-of established history on that, but I’m not sure.

src/encoded/static/components/motifs.js
```diff
@@ -189,14 +189,14 @@ export class MotifElement extends React.Component {
...
+        const footprintsLabel = (footprintsLength > 1 || footprintsLength === 0) ? `Footprints (${footprintsLength})` : 'Footprint';
         
         const targetComponent = (
             <React.Fragment>
             {targetList.length > 0 && (
-                <React.Fragment>
-                    {targetListLabel} {targetList}
-                </React.Fragment>
+                <div className='label-value-pair'>
+                    {targetListLabel} &nbsp; <div className='value'>{targetList}</div>
```
forresttanaka: It’s not wrong, but I personally prefer to use margins instead of non-breaking spaces to space elements. Since `targetListLabel` isn’t in its own div or span, I realize my way gets a little awkward.

forresttanaka (approved):
I like your changes. One small note that I don’t think is worth changing; just noting.

src/encoded/static/components/motifs.js
```diff
@@ -285,6 +352,60 @@ export const Motifs = (props) => {
...
+
+            if (d3lib) {
+                sequenceLogos.entryPoint(refToReference.current, fakePWM, d3lib, alignmentCoordinate, alignedStartCoordinate, referenceStart, '+', true);
+            }
+
+            return () => {
+                document.removeEventListener("scroll", trackScrolling);
+            }
+        });
+    }, [d3lib, props]);
```
forresttanaka: I see now. `props` is an object owned by React so that gets tricky in dependency arrays. But you can use the properties you use _inside_ of `props`. So `limit`, `referenceSequence`, `refToReference.current`, `alignmentCoordinate`, `alignedStartCoordinate`, `referenceStart`, `trackScrolling` (a function, but it gets reallocated on every render; you could make it unchanging by wrapping that function in `useCallback()`). ESLint might want `setD3Lib` as well, though it has no chance of changing since it’s a state-setting function.

src/encoded/static/components/motifs.js
```diff
@@ -285,6 +352,60 @@ export const Motifs = (props) => {
...
+    React.useEffect(() => {
+        if (limit === 0) {
+            document.addEventListener('scroll', trackScrolling);
+        }
+
+        // drawing reference sequence as a pwm so it lines up with the others
+        require.ensure(['d3', 'd3-sequence-logo'], (require) => {
+            setD3Lib(require('d3'));
+            const sequenceLogos = logos; // This is for local development when changes are needed to d3-sequence-logo.
+            // this.sequenceLogos = require('d3-sequence-logo');
```
forresttanaka: Can this be removed?

src/encoded/static/components/motifs.js
```diff
@@ -285,6 +352,60 @@ export const Motifs = (props) => {
...
+                if (nucleotide === 'A') {
+                    fakePWM.push([maxHeight, 0, 0, 0]);
+                } else if (nucleotide === 'C') {
+                    fakePWM.push([0, maxHeight, 0, 0]);
+                } else if (nucleotide === 'G') {
+                    fakePWM.push([0, 0, maxHeight, 0]);
+                } else {
+                    fakePWM.push([0, 0, 0, maxHeight]);
+                }
+            });
```
forresttanaka, line 395: I bet you can use a `.map()` here which might look more clear, so `const fakePWM = refSeq.map((nucleotide) => {...`. But `map`s an alternative, not a correction. I try to use it instead of `forEach` when possible, and it looks like `fakePWM` and `refSeq` have a one-to-one mapping, so I think it’s possible here.

forresttanaka (changes requested):
Just the dependency array needs more specific dependencies. The package.json includes the ESLint plugins, including the React one, so not sure why it didn’t give you its own list of dependencies.

src/encoded/static/components/motifs.js
dougli1sqrd: like just directly? Or do I go: `props.limit` etc

forresttanaka: In this case, I think it’s just `props.limit`. Every other dependency is a local variable. That’s one reason I destructure in the function declaration so that I don’t have `props` at all.

src/encoded/static/components/motifs.js
```diff
@@ -285,6 +351,59 @@ export const Motifs = (props) => {
...
+
+            if (d3lib) {
+                sequenceLogos.entryPoint(refToReference.current, fakePWM, d3lib, alignmentCoordinate, alignedStartCoordinate, referenceStart, '+', true);
+            }
+
+            return () => {
+                document.removeEventListener("scroll", trackScrolling);
+            }
+        });
+    }, [d3lib, props.limit, props.classList, props.context]);
```
forresttanaka: You’d normally use the variables you actually use in the `useEffect` callback, so since you assign `+props.limit` to limit, you can just use `limit` here. It doesn’t look like you use `classList` or `context` at all in the callback, so you don’t need those. You do need `refSequence`, `d3lib`, `refToReference`, `alignmentCoordinates`, `alignedStartCoordinate`, `referenceStart`. You might also need `trackScrolling` if ESLint complains. You also can’t use objects in the dependencies unless you want to check that you have a new copy of the object. The React dependency checks are very simple minded. It won’t detect that a value within an object has changed.

forresttanaka (changes requested):
Just need the dependency list completed, and a couple unused dependencies removed.

forresttanaka (approved):
OK, I think that should work. Just have to be careful that `trackScrolling` doesn’t cause this callback to get called needlessly. Can only tell that in testing.

src/encoded/static/components/motifs.js
```diff
@@ -87,8 +87,7 @@ export class MotifElement extends React.Component {
     componentDidMount() {
         require.ensure(['d3', 'd3-sequence-logo'], (require) => {
             this.d3 = require('d3');
-            // this.sequenceLogos = logos; // This is for local development when changes are needed to d3-sequence-logo.
-            this.sequenceLogos = require('d3-sequence-logo');
+            this.sequenceLogos = logos; // This is for local development when changes are needed to d3-sequence-logo.
```
hitz: unrelated to issues but seems like this should be put back to the require before we merge.  Do we have to build the d3-sequence-logo package?

forresttanaka (approved):

forresttanaka (approved):
OK, I sure hope that’s OK. I wish I could tell you more about it, but at this stage you know at least as much as I do.

PR #152, opened 2022-08-17
REGD-64-allele-info

mingjiecn:
I changed the genomic_data_service_url so regulome app can point to the new database with all allele info. Will change it back before merge.

### DIFF
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -939,7 +939,6 @@ export class RegulomeSearch extends React.Component {
                 if (snp.chrom === chrom && snp.coordinates.gte === +start && snp.coordinates.lt === +end) {
                     hitSnps[snp.rsid] = {};
                     const populationAlleles = {};
-
                     if (snp.ref_allele_freq) {
                         const refAlleleTag = Object.keys(snp.ref_allele_freq)[0];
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/regulome_search.js
```diff
@@ -938,24 +939,45 @@ export class RegulomeSearch extends React.Component {
...
+                                Object.keys(snp.ref_allele_freq[allele]).forEach((population) => {
+                                    if (!snp.ref_allele_freq[allele][population]) {
+                                        populationAlleles[population] = [`${allele}=N/A`]
+                                    } else {
+                                        populationAlleles[population] = [`${allele}=${snp.ref_allele_freq[allele][population]}`];
+                                    }
+                                    
+                                        
+                                    
+                                });
```
forresttanaka, line 952: Not specifically asking for a change, but I think the next person to modify this code would likely delete these blank lines because you don’t often find blank lines before the ends of code blocks. I think ESLint would mark these as errors, though I thought regulome used ESLint so maybe I’m wrong about that. The same thing would apply to some blank lines at the ends of code blocks below.

forresttanaka (approved):
My only minor concern are some blank lines in unusual places.

src/encoded/static/components/regulome_search.js
mingjiecn: I didn't use ESlint. Will use it and fix the format.

PR #153, opened 2022-08-25
Regd 58 split qtl table

shengchd:
- Split QTL table by methods (caQTLs vs eQTLs)
- Added population column and dropped targets column for caQTLs
- Dropped targets column under QTL thumbnail

### DIFF
--- a/ini-templates/development-template.ini
+++ b/ini-templates/development-template.ini
@@ -9,7 +9,7 @@ pyramid.debug_authorization = false
 pyramid.debug_notfound = true
 pyramid.debug_routematch = false
 
-genomic_data_service_url = http://34.217.63.100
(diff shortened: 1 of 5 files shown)


ini-templates/development-template.ini
```diff
@@ -9,7 +9,7 @@ pyramid.debug_authorization = false
 pyramid.debug_notfound = true
 pyramid.debug_routematch = false
 
-genomic_data_service_url = http://34.217.63.100
+genomic_data_service_url = http://52.42.245.234
```
hitz: Not sure this is the correct server but if not we will have to change it before we merge.

hitz (approved):
Looks Good!

PR #167, opened 2022-11-08
Regd-81-remove-overlapping-hits

shengchd:
- filtered overlapping peaks from the same datasets in ChIP, DNase & ATAC-seq
- corrected the total peak number after filtering

### DIFF
--- a/src/encoded/static/components/regulome_search.js
+++ b/src/encoded/static/components/regulome_search.js
@@ -545,29 +545,29 @@ const populationOrder = [
 ];
 
 // function to de-dup overlapping peaks in each dataset from ChIP, DNase and ATAC-seq assays
-function filter_overlapping_peaks(Datasets) {
-    var Datasets_filtered = [];
-    var Datasets_filtered_count = 0;
+function filterOverlappingPeaks(Datasets) {
+    const DatasetsFiltered = [];
+    let DatasetsFilteredCount = 0;
     if (Datasets.length > 0) {
-        //we want to keep the wider peaks in each dataset
-        //if the start & end positions are the same we will keep the one with the strongest signal
-        //sort all peaks by dataset ids -> start positions in ascending order -> end positions in descending order -> signals in descending order
-        Datasets.sort((a,b)=>a.dataset_rel.localeCompare(b.dataset_rel) || a.start-b.start || b.end-a.end || b.value-a.value);
-        //keep the peak if it is the first peak in a new dataset or if it is not within the previous peak
-        Datasets_filtered.push(Datasets[0]);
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/regulome_search.js
```diff
@@ -544,6 +544,32 @@ const populationOrder = [
...
 
+// function to de-dup overlapping peaks in each dataset from ChIP, DNase and ATAC-seq assays
+function filter_overlapping_peaks(Datasets) {
+    var Datasets_filtered = [];
+    var Datasets_filtered_count = 0;
+    if (Datasets.length > 0) {
+        //we want to keep the wider peaks in each dataset
+        //if the start & end positions are the same we will keep the one with the strongest signal
+        //sort all peaks by dataset ids -> start positions in ascending order -> end positions in descending order -> signals in descending order
+        Datasets.sort((a,b)=>a.dataset_rel.localeCompare(b.dataset_rel) || a.start-b.start || b.end-a.end || b.value-a.value);
```
mingjiecn: value is not used for sorting later, do we still need to sort by value?

src/encoded/static/components/regulome_search.js
```diff
@@ -544,6 +544,32 @@ const populationOrder = [
...
+            if (Datasets[i].dataset_rel !== lastDataset || Datasets[i].end > lastEnd) { 
+                Datasets_filtered.push(Datasets[i]); 
+                lastDataset = Datasets[i].dataset_rel;
+                lastEnd = Datasets[i].end;
+            } else {
+                Datasets_filtered_count ++;
+            }
+        }
+    }
+    return [Datasets_filtered,Datasets_filtered_count];
```
mingjiecn: Should have one space after comma. Did you lint your code before pushing?

mingjiecn (changes requested):
some minor concerns need to be addressed

src/encoded/static/components/regulome_search.js
shengchd: We are showing signals (i.e. values) for ChIP-seq peaks, so I thought it's better to explicitly keep the one with the strongest signal this way (when there are multiple peaks with exactly same positions from the same dataset)?

src/encoded/static/components/regulome_search.js
shengchd: Thanks for pointing out that, my eslint wasn't working correctly. I have also fixed other stylistic issues in my code.

src/encoded/static/components/regulome_search.js
mingjiecn: sound good!

mingjiecn (approved):

PR #178, opened 2023-03-07
REGD-88 replace new download file

shengchd:
Replaced the download file with the new one calculated in GRCh38


emro (approved):
Looks great!

